PR #71, opened 2018-04-11
Update interface to fcu

mrivi:
Enables:
- sending OBSTACLE_DISTANCE message to the FCU
- receiving a planned path from the FCU e.g. mission from `/mavros/trajectory/desired`
- sending a collision free path to the FCU with `/mavros/trajectory/generated` in any mode
- still support old offboard interface

To do:
- [ ] ~~send message to FCU when a waypoint is not reachable~~
- [x] send obstacle_distance to the FCU even when the local planner doesn't calculate the polar histogram e.g. backoff from obstacles

### DIFF
--- a/local_planner/CMakeLists.txt
+++ b/local_planner/CMakeLists.txt
@@ -12,7 +12,10 @@ find_package(catkin REQUIRED COMPONENTS
   dynamic_reconfigure
   tf
   pcl_ros
+  mavros
+  mavros_extras
   mavros_msgs
+  mavlink
 )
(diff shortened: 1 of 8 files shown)


LorenzMeier:
@mrivi Could you reference the firmware PR and help get both in? Please ping @baumanta and me regarding tests. You're more than welcome to join them!

mrivi:
@LorenzMeier sure! 

This is the Firmware PR https://github.com/PX4/Firmware/pull/9270

nicovanduijn:
I can confirm that it works as expected in simulation

PR #76, opened 2018-06-04
Global Planner: update obstacle avoidance interface

mrivi:
This PR brings the new obstacle avoidance interface on the global planner

### DIFF
--- a/global_planner/CMakeLists.txt
+++ b/global_planner/CMakeLists.txt
@@ -14,7 +14,10 @@ find_package(catkin REQUIRED COMPONENTS
   message_generation
   tf
   pcl_ros
-  octomap_msgs
+  mavlink
+  mavros_msgs
+  mavros
+  mavros_extras
 )
 find_package(PCL 1.7 REQUIRED)
 find_package(octomap REQUIRED)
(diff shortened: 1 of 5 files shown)


nicovanduijn:
I tried this, but somehow it never advances to the first mission item after takeoff. Similar to the corner case we found with the local planner, but more consistent. Do you experience this too?

mrivi:
I'll look into it

mhkabir:
Hey guys, @nicovanduijn @mrivi. Do either of you plan to have a look at this soon?

nicovanduijn:
I've been using it in SITL and it works just fine. I'm not doing any actual flight testing these days, so I can't do that.

Not sure if this is a priority for @mrivi or @baumanta, but it would of course be nice to support the global planner again at some point

mhkabir:
Is the mission getting stuck issue still present with the latest Firmwares changes?

mrivi:
@mhkabir no, it's solved

LorenzMeier:
@baumanta @mrivi What is the status of this PR?

mrivi:
@LorenzMeier it has been used in simulation by @nicovanduijn for his thesis but it has never been tested on hardware

nicovanduijn:
I think we should merge this. 

The global planner hasn't been flight-tested in so long that without this, it won't even be testable (using the offboard & RViz is just a pain), and this pr at least brings it into a state that works in SITL and is somewhat in sync with the local planner in terms of the interface.

LGTM

PR #78, opened 2018-06-06
Fix interface

mrivi:
- set yaw waypoint to current yaw when rising to goal altitude
- update goal from FCU only when in mission
- when updating goal from FCU check difference with respect to old goal also in z coordinate. Fixes problem with the first takeoff waypoint

### DIFF
--- a/local_planner/src/nodes/histogram.cpp
+++ b/local_planner/src/nodes/histogram.cpp
@@ -122,17 +122,17 @@ void Histogram::setZero() {
   }
 }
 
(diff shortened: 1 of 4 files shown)


LorenzMeier:
@mrivi Thanks! Feel free to assign @baumanta for reviews so she gets notified.

mrivi:
Logs from test https://logs.px4.io/plot_app?log=71898feb-dba9-4824-be9f-d378dbce7313

Issue #79, opened 2018-06-09
bezier.h:132:12: error: return-statement with a value, in function returning 'void' [-fpermissive]


ArkadiuszNiemiec:
I am trying to build your node in ROS Melodic on Ubuntu 18.04 and it fails with:
`/home/arek/catkin_ws/src/avoidance/global_planner/include/global_planner/bezier.h:132:12: error: return-statement with a value, in function returning 'void' [-fpermissive]
     return path;
`
Problematic line:
https://github.com/PX4/avoidance/blob/acebb50ba6d7d0cb8dc9a99aeddc7b9e8469485a/global_planner/include/global_planner/bezier.h#L132

I have pushed a fix just to build it but I assume it's not how the code is intended to work. I will try to push more commits if I will make the whole package work and test it.

ArkadiuszNiemiec:
I have tested the `global_planner` with #80 and it worked fine. I am closing this issue for now.

PR #80, opened 2018-06-09
Fix return error

ArkadiuszNiemiec:
Simple fix for the issue #79


ArkadiuszNiemiec:
I have pushed a fix just to build it but I assume it's not how the code is intended to work. I will try to push more commits if I will make the whole package work and test it.

PR #89, opened 2018-07-06
Introduce yaw goal

nicovanduijn:
For the navigator to check off a mission item, the avoidance needs to be able to assume a defined yaw once it has reached the goal radius. This PR adds functionality for the local planner to do so

### DIFF
--- a/local_planner/src/nodes/star_planner.cpp
+++ b/local_planner/src/nodes/star_planner.cpp
@@ -33,8 +33,8 @@ void StarPlanner::setBoxSize(Box histogram_box_size) {
   histogram_box_size_ = histogram_box_size;
 }
 
-void StarPlanner::setGoal(geometry_msgs::Point goal) {
-  goal_ = goal;
+void StarPlanner::setGoal(geometry_msgs::PoseStamped goal) {
+  goal_ = goal.pose.position;
   tree_age_ = 1000;
 }
 
--- a/local_planner/src/nodes/local_planner.cpp
+++ b/local_planner/src/nodes/local_planner.cpp
@@ -98,7 +98,7 @@ void LocalPlanner::logData() {
             << reach_altitude_ << "\t" << use_ground_detection_ << "\t"
             << obstacle_ << "\t" << height_change_cost_param_adapted_ << "\t"
             << over_obstacle_ << "\t" << too_low_ << "\t" << is_near_min_height_
-            << "\t" << goal_.x << "\t" << goal_.y << "\t" << goal_.z << "\t"
+            << "\t" << goal_.pose.position.x << "\t" << goal_.pose.position.y << "\t" << goal_.pose.position.z << "\t"
             << algorithm_total_time_[algorithm_total_time_.size() - 1] << "\t"
             << tree_time_[tree_time_.size() - 1] << "\n";
     myfile1.close();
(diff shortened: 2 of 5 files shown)


local_planner/src/nodes/star_planner.cpp
```diff
@@ -33,8 +33,8 @@ void StarPlanner::setBoxSize(Box histogram_box_size) {
   histogram_box_size_ = histogram_box_size;
 }
 
-void StarPlanner::setGoal(geometry_msgs::Point goal) {
-  goal_ = goal;
+void StarPlanner::setGoal(geometry_msgs::PoseStamped goal) {
```
thomasgubler: are we passing by value on purpose here?

nicovanduijn: good catch

nicovanduijn:
My investigations have yielded a few results.

Currently the desired yaw received from PX4 (/mavros/trajectory/desired/point_2.yaw) does not contain very meaningful information. It currently seems to encode the heading toward the goal, which is redundant information since we are already sending current position in point_1 and goal location in point_2.

Therefore I suggest one of the following:
- We change the yaw sent to the avoidance to be the one the navigator actually checks for: the straight connection line between mission points. This makes little practical sense I think.

- Change the logic in is_mission_item_reached() to ignore yaw. I don't know enough about the navigator in order to do this with confidence that I won't break any other logic (e.g. how does this affect fixed-wing and vtol missions??). I'm happy to be educated on this part, a first naive attempt is here: https://github.com/nicovanduijn/Firmware/commit/f24ffffc6d4eac8659838d987038782bc01a7de0
This would of course render this PR useless, which I'm okay with.

- Change the logic in is_mission_item_reached() to check for the yaw defined by the mission item, unless it is NAN. (Maybe use the _mission_item.force_heading boolean for this? ) This might be more involved, but would make the most sense I think. It would also add the feature of being able to enforce a specific orientation at a certain waypoint, which could be useful (think: inspection)

Any thoughts?

mrivi:
The interface lies in the mc_pos_control and it forwards the `setpoint_position_triplet` from navigator. The yaw is continuously updated by navigator here https://github.com/PX4/Firmware/blob/master/src/modules/navigator/mission.cpp#L1171 . 
To be able to send the desired yaw at a waypoint to the avoidance navigator needs to change. I remember that @stifael proposed to get rid of the continuos update from navigator a while back. I don't remember what was the issue with the proposal.

I am in favor of not checking for yaw at waypoints when the avoidance is running especially now that we are limited by a single forward facing camera. Avoidance should be setting the heading. 

FYI - @dagar thoughts?

dagar:
I'm in favor of removing the continuous yaw updates from navigator. It needlessly increases the load and logging rate.

Other than conflicting with current FlightTasks mc_pos_control work I don't think it's even a difficult change to make.

Stifael:
For mc, this PR https://github.com/PX4/Firmware/pull/9146 removes the continuous update for mc.

But it will create again conflict with the mc_pos_cpntrol_main changes of flighttask.

mrivi:
Sure, I understand. We could do it on top of the branch that has the avoidance interface for flight task.

nicovanduijn:
I see it makes sense for this to wait - I'll put the commit for passing by const& on a seperate PR as it has nothing to do with this and should get merged regardless of how we proceed here.

thomasgubler:
From dev call:
@mrivi to look into options in navigator to check if we are in avoidance mode

mrivi:
Draft of the proposal made during the dev call: https://github.com/PX4/Firmware/pull/9880

If the avoidance is active and healthy, yaw is never checked.

thomasgubler:
@mrivi can we close this since https://github.com/PX4/Firmware/pull/10037 got merged ?

mrivi:
Yes, we can close this but the PR implementing the actual feedback is this one https://github.com/PX4/Firmware/pull/10413

PR #92, opened 2018-07-17
Architecture change rebased

baumanta:
Change in avoidance architecture. The planner is now completely decoupled from the waypoint generation. Therefore the waypoints can be published at a constant rate. Also both position and velocity setpoints are supported and it can be chosen which to use from rqt reconfigure

Tested in SITL and on one flight with Intel Aero

### DIFF
--- a/local_planner/src/nodes/common.cpp
+++ b/local_planner/src/nodes/common.cpp
@@ -142,6 +142,14 @@ double velocitySigmoid(double max_vel, double min_vel, double slope, double v_ol
   return speed;
 }
 
+double velocityLinear(double max_vel, double min_vel, double slope, double v_old, double elapsed) {
+  v_old -= min_vel;
+  double t_old = v_old/slope;
+  double t_new = t_old + elapsed;
+  double speed = min_vel + t_new*slope;
+  return speed;
+}
+
 double getAngularVelocity(double new_yaw, double curr_yaw) {
   while (new_yaw > M_PI){
 	  new_yaw -= M_PI;
@@ -171,6 +179,6 @@ double getAngularVelocity(double new_yaw, double curr_yaw) {
 	  vel = yaw_vel2;
   }
 
-  return 0.3 * vel * std::abs(vel);
+  return vel;
 
 }
--- a/local_planner/src/nodes/common.cpp
+++ b/local_planner/src/nodes/common.cpp
@@ -133,8 +133,8 @@ void normalize(geometry_msgs::Point &p) {
 }
 
 double velocitySigmoid(double max_vel, double min_vel, double slope, double v_old, double elapsed) {
-  max_vel+=0.01;
-  min_vel-=0.01;
+  max_vel+=0.05;
+  min_vel-=0.05;
   v_old -= min_vel;
   double t_old = - 1.0/slope *log((max_vel - min_vel)/v_old - 1.0);
   double t_new = t_old + elapsed;
--- a/local_planner/src/nodes/common.cpp
+++ b/local_planner/src/nodes/common.cpp
@@ -132,53 +132,54 @@ void normalize(geometry_msgs::Point &p) {
   }
 }
 
-double velocitySigmoid(double max_vel, double min_vel, double slope, double v_old, double elapsed) {
-  max_vel+=0.05;
-  min_vel-=0.05;
+double velocitySigmoid(double max_vel, double min_vel, double slope,
+                       double v_old, double elapsed) {
+  max_vel += 0.05;
+  min_vel -= 0.05;
   v_old -= min_vel;
-  double t_old = - 1.0/slope *log((max_vel - min_vel)/v_old - 1.0);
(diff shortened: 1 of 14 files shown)


nicovanduijn:
Not a proper review, but here are a few first impressions:
- Some more comments would be nice. This is all still quite hard to understand if one isn't too familiar with the code. Maybe drop a quick line in the header files on each function, describe in and outputs and what it does. Also some of the member variables should be documented if their name doesn't make it abundantly clear what they are. (What is pose_ ? How is it different from offboard_pose_?)
- I've found quite a few instances where you can pass by (const) reference. Generally, anything larger than 32 bits makes sense to pass by reference. Personally, I feel like it's okay for doubles, as it may not even be faster depending on the compiler optimizations. But if you're passing a geometry_msgs::PoseStamped that contains like 5 doubles, you should definitely pass by reference.
- Some of these files include a lot of other stuff (waypoint_generator.h). Can you double-check if they really are all necessary? Maybe some have become obsolete in the meantime

local_planner/src/nodes/common.cpp
```diff
@@ -99,22 +99,22 @@ int azimuthAngletoIndex(int z, int res) {  //[-180,180]
 }
 
 // calculate the yaw for the next waypoint
-double nextYaw(geometry_msgs::PoseStamped u, geometry_msgs::Vector3Stamped v,
+double nextYaw(geometry_msgs::PoseStamped u, geometry_msgs::Point v,
                double last_yaw_) {
```
nicovanduijn: I assume last_yaw_ is some legacy piece?

local_planner/src/nodes/common.cpp
```diff
@@ -131,3 +131,55 @@ void normalize(geometry_msgs::Point &p) {
...
+  v_old -= min_vel;
+  double t_old = v_old / slope;
+  double t_new = t_old + elapsed;
+  double speed = min_vel + t_new * slope;
+  return speed;
+}
+
+double getAngularVelocity(double new_yaw, double curr_yaw) {
+  while (new_yaw > M_PI) {
+    new_yaw -= M_PI;
```
nicovanduijn: Is this correct? Why not `new_yaw -= 2*M_PI` ? I haven't completely thought it through yet, but conditionally rotating by 180 degrees is counter-intuitive at best. (In that case, clearly explain with comments why it is done)

Maybe refactor this logic into a dedicated function like `double wrapToPi(const double& input)`

local_planner/src/nodes/common.cpp
```diff
@@ -131,3 +131,55 @@ void normalize(geometry_msgs::Point &p) {
...
+double velocityLinear(double max_vel, double min_vel, double slope,
+                      double v_old, double elapsed) {
+  v_old -= min_vel;
+  double t_old = v_old / slope;
+  double t_new = t_old + elapsed;
+  double speed = min_vel + t_new * slope;
+  return speed;
+}
+
+double getAngularVelocity(double new_yaw, double curr_yaw) {
```
nicovanduijn: If I understand it right, this function computes a good yaw rate to reach a given desired orientation and the current orientation, using some hard-coded p-gain. The function and parameter naming is a bit misleading: I could easily think that this derives an angular velocity based on two orientation samples. Also `new_yaw` makes it sound like an output. Maybe `desired_yaw` would more clearly indicate what the use of this quantity is.

local_planner/src/nodes/local_planner_node.cpp
```diff
@@ -536,6 +541,20 @@ void LocalPlannerNode::publishWaypoints() {
...
   adapted_wp_pub_.publish(sphere2);
   smoothed_wp_pub_.publish(sphere3);
+  waypoint_pub_.publish(result.path);
+  path_pub_.publish(path_actual_);
+  publishSetpoint(result.velocity_waypoint, result.waypoint_type);
+
+  // to mavros
+  mavros_msgs::Trajectory obst_free_path = {};
+  if(local_planner_.use_vel_setpoints_){
+      mavros_vel_setpoint_pub_.publish(result.velocity_waypoint);
```
mrivi: I think the transformation from `world` to `local_origin` is missing.

baumanta: As discussed on slack: I changed all the frames to world frame to have a consistent use of frames

local_planner/src/nodes/common.cpp
```diff
@@ -99,22 +99,22 @@ int azimuthAngletoIndex(int z, int res) {  //[-180,180]
...
 
   return atan2(dy, dx);
 }
 
-geometry_msgs::PoseStamped createPoseMsg(geometry_msgs::Vector3Stamped waypt,
+geometry_msgs::PoseStamped createPoseMsg(geometry_msgs::Point waypt,
                                          double yaw) {
   geometry_msgs::PoseStamped pose_msg;
   pose_msg.header.stamp = ros::Time::now();
   pose_msg.header.frame_id = "/world";
```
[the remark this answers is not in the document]
baumanta: oh ok, I think I got them all now

PR #98, opened 2018-08-16
launch scripts: fixes to launch PX4

bkueng:
The upstream PX4 CLI interface got changed. This changes to use the newly
added px4.launch script, which makes the avoidance repo independent from
the CLI interface.

Requires https://github.com/PX4/Firmware/pull/10247


bkueng:
That was quick, thanks!

PR #106, opened 2018-09-12
Bug fix and some improvements

baumanta:
- fix bug in smoothing, where the last time was not always saved which led to incorrect time differentials
- log cpu and memory usage in rostopics
- change markers for actual path and path directed by the waypoints. These markers before accumulated over time. This led to message bandwidths of 160MB/s after a few minutes for each path and overloaded the memory when logging.
- set Realsense visual preset to HIGH_ACCURACY mode by default as it proved to be better than the default mode

We got nice flights with these changes

### DIFF
--- a/local_planner/src/nodes/local_planner_node.cpp
+++ b/local_planner/src/nodes/local_planner_node.cpp
@@ -80,7 +80,8 @@ LocalPlannerNode::LocalPlannerNode() {
       nh_.advertise<sensor_msgs::LaserScan>("/mavros/obstacle/send", 10);
   current_waypoint_pub_ =
       nh_.advertise<visualization_msgs::Marker>("/current_setpoint", 1);
-  log_name_pub_ = nh_.advertise<std_msgs::String>("/log_name", 1);
+  cpu_pub_ = nh_.advertise<std_msgs::String>("/cpu_usage", 1);
+  memory_pub_ = nh_.advertise<std_msgs::String>("/memory_usage", 1);
   takeoff_pose_pub_ =
       nh_.advertise<visualization_msgs::Marker>("/take_off_pose", 1);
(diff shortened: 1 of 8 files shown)


local_planner/src/nodes/waypoint_generator.cpp
```diff
@@ -69,6 +69,10 @@ void WaypointGenerator::calculateWaypoint() {
     }
   }
   last_wp_type_ = planner_info_.waypoint_type;
+  last_time_ = ros::Time::now();
```
bkueng: This should use the same timestamp as used in the current timestep (`now` in `getPathMsg`), which means that `now` should be set in here, and then passed as argument to `getPathMsg`.
This makes sure there is no time slip due to computation time in between.

local_planner/src/nodes/local_planner_node.cpp
```diff
@@ -908,10 +946,17 @@ void LocalPlannerNode::threadFunction() {
...
     }
 
-    // publish log name
-    std_msgs::String msg;
-    msg.data = local_planner_.log_name_;
-    log_name_pub_.publish(msg);
+    // publish performance data
+    std_msgs::String cpu_msg;
+    std_msgs::String mem_msg;
+    std::string cpu = exec(" top -bn 1 | awk '{print $9}' | tail -n +8 | awk '{s+=$1} END {print s/8}'");
```
bkueng: CPU & Memory monitoring should IMO not be part of the planner node.
There are already existing system monitoring nodes for ROS, for example https://github.com/ethz-asl/ros-system-monitor. Is there any reason not to use that?

baumanta: sounds good, I will look into that

PR #108, opened 2018-09-20
Launch files: world file name argument

mrivi:
create arg with the world file name such that to change the simulation world only one argument needs to be edited


baumanta:
thanks that does look better!

PR #110, opened 2018-09-25
Dev 3cam rebased

baumanta:
Changes to the algorithm which enable us to fly with up to 3 cameras.


baumanta:
this branch has been flight tested on 20.09.18.

local_planner/src/nodes/local_planner_node.cpp
```diff
@@ -739,8 +767,11 @@ void LocalPlannerNode::printPointInfo(double x, double y, double z) {
   printf("-------------------------------------------- \n");
 }
 
-void LocalPlannerNode::pointCloudCallback(const sensor_msgs::PointCloud2& msg) {
-  newest_point_cloud_ = msg; // FIXME: avoid a copy
+void LocalPlannerNode::pointCloudCallback(const sensor_msgs::PointCloud2::ConstPtr& msg, int index) {
+
+  available_clouds_[index] = true;
```
mrivi: I think this booleans are never reset to false

local_planner/cfg/local_planner_node.cfg
```diff
@@ -30,26 +30,26 @@ gen.add("min_realsense_dist_", double_t, 0, "Discard points closer than that", 0
 gen.add("min_plane_percentage_", double_t, 0, "Discard planefits with less inliers", 0.7, 0, 1)
 gen.add("avoid_radius_", double_t, 0, "min dist before absolutely avoiding", 2.5, 0, 10)
 gen.add("min_dist_backoff_", double_t, 0, "min dist before backing off", 1.5, 0, 10)
-gen.add("pointcloud_timeout_hover_", double_t, 0, "after this timeout the controller sends hover commands", 0.25, 0, 10)
+gen.add("pointcloud_timeout_hover_", double_t, 0, "after this timeout the controller sends hover commands", 0.5, 0, 10)
 gen.add("pointcloud_timeout_land_", double_t, 0, "after this timeout the controller sends hover commands", 15, 0, 1000)
-gen.add("reproj_age_", int_t, 0, "maximum age of a reprojected point", 50, 0, 1000)
+gen.add("reproj_age_", int_t, 0, "maximum age of a reprojected point", 0, 0, 1000)
```
bkueng: Does this change aply to all configs? I.e. do we want to disable it for single cams as well?

local_planner/src/nodes/local_planner_node.cpp
```diff
@@ -110,24 +119,43 @@ void LocalPlannerNode::readParams() {
...
+  for(int i=0; i<available_clouds_.size(); ++i){
+	  if(available_clouds_[i] == true){
+		  if(!tf_listener_.canTransform("/local_origin", newest_cloud_msgs_[i].header.frame_id,
+				  newest_cloud_msgs_[i].header.stamp)){
+			  missing_transforms ++;
+		  }
+	  }
+  }
+
+  if(missing_transforms == 0){
```
bkueng: `return missing_transforms == 0;`

local_planner/src/nodes/local_planner.h
```diff
@@ -228,7 +228,7 @@ class LocalPlanner {
   geometry_msgs::PoseStamped take_off_pose_;
   geometry_msgs::PoseStamped offboard_pose_;
   sensor_msgs::LaserScan distance_data_ = {};
-  pcl::PointCloud<pcl::PointXYZ> complete_cloud_;
+  std::vector<pcl::PointCloud<pcl::PointXYZ>> complete_cloud_;
```
bkueng: Can you comment that this contains the point clouds from N cameras?

local_planner/src/nodes/local_planner_node.cpp
```diff
@@ -3,15 +3,20 @@
...
 
   // Set up Dynamic Reconfigure Server
   dynamic_reconfigure::Server<avoidance::LocalPlannerNodeConfig>::CallbackType
       f;
   f = boost::bind(&LocalPlannerNode::dynamicReconfigureCallback, this, _1, _2);
   server_.setCallback(f);
 
-  pointcloud_sub_ = nh_.subscribe<const sensor_msgs::PointCloud2&>(
-      depth_points_topic_, 1, &LocalPlannerNode::pointCloudCallback, this);
+  pointcloud_front_sub_ = nh_.subscribe<sensor_msgs::PointCloud2>(
```
bkueng: I'd like this to be more generic:
- Add a struct with all data for a single camera (point cloud subscription, newest_cloud_msg, ...)
- Then depending on the configured cameras (check `nh_.param<std::string>("depth_points_topic_front", depth_points_topic_front_,  "/camera_front/depth/points");`, initialize a vector of that struct and do the subscriptions.
- Then pass the number of cameras to all classes that need to know it and initialize the vectors for example with `(num_cameras, false)`, instead of the hardcoded size `{false, false, false}`.

local_planner/src/nodes/local_planner_node.cpp
baumanta: but with the arbitrary camera number this variable has been eliminated

local_planner/src/nodes/local_planner_node.cpp
```diff
@@ -978,27 +1026,44 @@ int main(int argc, char **argv) {
...
     } else {
-      if (Node.never_run_) {
+      if (Node.never_run_ || (since_last_cloud > pointcloud_timeout_hover &&
+                              since_start > pointcloud_timeout_hover)) {
         if (Node.position_received_) {
           hover = true;
+          std::string not_received = ", no cloud received on topic ";
+          for (int i = 0; i < Node.cameras_.size(); i++) {
+            if (!Node.cameras_[i].received_) {
+              not_received.append(Node.cameras_[i].topic_);
```
bkueng: Can be simplified to: `not_received += Node.cameras_[i].topic_ + " ";`

local_planner/cfg/local_planner_node.cfg
bkueng: I don't know what the proper way of doing that in ROS would be. Your solution sounds fine.

PR #111, opened 2018-10-02
clean up - smaller fixes

baumanta:
Delete all code related to ground detection. This is an over engineered solution of an easy problem. This code was not used in a while (and will probably never be used anymore). Also is is quite hard to maintain.

Change the shape of the cropped pointcloud. Before it was a rectangular box. Now it is a sphere, which is cut off at the bottom and the top. This is a more natural shape for the polar histogram and can be achieved without additional computations.

bug fix: starting position was often not updated which resulted in a sideways motion once the drone is switched to an auto mode. The desired behavior would be a straight upwards motion.

feature: reducing the velocity before the goal is reached to avoid a hard stop and overshoot.


baumanta:
this PR was flight tested on 02.10.18 and showed the desired behavior.

LorenzMeier:
Awesome! Deleting code is always great!

local_planner/src/nodes/waypoint_generator.cpp
```diff
@@ -339,6 +307,21 @@ void WaypointGenerator::adaptSpeed() {
   double delta_dist = since_update_sec * curr_vel_magnitude_;
   speed_ += delta_dist;
 
+  //break before goal
+  geometry_msgs::Point a;
+  a.x = std::abs(goal_.x - pose_.pose.position.x);
+  a.y = std::abs(goal_.y - pose_.pose.position.y);
+  a.z = std::abs(goal_.z - pose_.pose.position.z);
+  if (a.x < 3 * curr_vel_magnitude_ && a.y < 3 * curr_vel_magnitude_) {
```
mrivi: if either `a.x` or `a.y` are `>= 3.0 * curr_vel_magnitude` and  `<= 4.0 * curr_vel_magnitude`, `limit_speed_close_to_goal_` is not updated. Is that intended?

local_planner/src/nodes/local_planner_node.cpp
```diff
@@ -494,65 +413,26 @@ void LocalPlannerNode::publishBox() {
...
   box.pose.orientation.y = 0.0;
   box.pose.orientation.z = 0.0;
   box.pose.orientation.w = 1.0;
-  box.scale.x = local_planner_.histogram_box_size_.xmax_ +
-                local_planner_.histogram_box_size_.xmin_;
-  box.scale.y = local_planner_.histogram_box_size_.ymax_ +
-                local_planner_.histogram_box_size_.ymin_;
-  box.scale.z = local_planner_.histogram_box_size_.zmax_ +
-                local_planner_.histogram_box_size_.zmin_;
+  box.scale.x = 2 * local_planner_.histogram_box_.radius_;
```
mrivi: `2` -> `2.0`

local_planner/src/nodes/local_planner.cpp
```diff
@@ -470,7 +405,7 @@ void LocalPlanner::reprojectPoints(Histogram histogram) {
           dist = distance3DCartesian(pose_.pose.position, temp_array[i]);
           age = histogram.get_age(e, z);
 
-          if (dist < 2 * histogram_box_size_.xmax_ && dist > 0.3 &&
+          if (dist < 2 * histogram_box_.radius_ && dist > 0.3 &&
```
mrivi: `2` -> `2.0`

local_planner/src/nodes/box.cpp
```diff
@@ -1,41 +1,25 @@
...
 
-Box::Box(double x_min, double x_max, double y_min, double y_max, double z_min,
-         double z_max)
-    : xmin_{x_min},
-      xmax_{x_max},
-      ymin_{y_min},
-      ymax_{y_max},
-      zmin_{z_min},
-      zmax_{z_max} {}
+Box::Box(double radius)
```
mrivi: can be `const`

local_planner/src/nodes/box.cpp
```diff
@@ -1,41 +1,25 @@
...
 
 // update bounding box limit coordinates around a new UAV pose
-void Box::setLimitsHistogramBox(geometry_msgs::Point pos, Box size) {
-  xmin_ = pos.x - size.xmin_;
-  ymin_ = pos.y - size.ymin_;
-  zmin_ = pos.z - size.zmin_;
-  xmax_ = pos.x + size.xmax_;
-  ymax_ = pos.y + size.ymax_;
-  zmax_ = pos.z + size.zmax_;
+void Box::setBoxLimits(geometry_msgs::Point pos) {
```
mrivi: can be `const`

local_planner/src/nodes/box.cpp
```diff
@@ -1,41 +1,25 @@
...
-  xmin_ = pos.x - size.xmin_;
-  ymin_ = pos.y - size.ymin_;
-  zmin_ = pos.z - min_dist - size.zmin_;
-  xmax_ = pos.x + size.xmax_;
-  ymax_ = pos.y + size.ymax_;
-  zmax_ = pos.z;
-}
-
-bool Box::isPointWithin(double x, double y, double z) {
+bool Box::isPointWithinBox(double x, double y, double z) {
```
mrivi: the doubles can be `const`

local_planner/src/nodes/waypoint_generator.cpp
```diff
@@ -339,6 +307,21 @@ void WaypointGenerator::adaptSpeed() {
   double delta_dist = since_update_sec * curr_vel_magnitude_;
   speed_ += delta_dist;
 
+  //break before goal
+  geometry_msgs::Point a;
```
mrivi: I would give a more meaningful name

local_planner/src/nodes/waypoint_generator.cpp
baumanta: yes, that is intended. It needs to have a hysteresis in there, because the distance at which we start to limit the velocity is velocity dependent. Meaning if we slow down the vehicle, the distance at which we should do so shrinks. This would lead to velocity discontiuities

local_planner/src/nodes/box.cpp
```diff
@@ -1,41 +1,25 @@
...
 
-Box::Box(double x_min, double x_max, double y_min, double y_max, double z_min,
-         double z_max)
-    : xmin_{x_min},
-      xmax_{x_max},
-      ymin_{y_min},
-      ymax_{y_max},
-      zmin_{z_min},
-      zmax_{z_max} {}
+Box::Box(const double radius)
```
nicovanduijn: Just stumbled on this; there's little added benefit in declaring const if you're passing by value. I believe what Martina meant was that you should generally pass by constant reference any value that is larger than 32 bits

baumanta:
@mrivi does that look good to you now? You think we can merge it?

PR #116, opened 2018-10-13
Remove mav_msgs dependency

mhkabir:
Unused dependency.

### DIFF
--- a/global_planner/CMakeLists.txt
+++ b/global_planner/CMakeLists.txt
@@ -112,7 +112,7 @@ generate_dynamic_reconfigure_options(
 catkin_package(
   INCLUDE_DIRS include
   LIBRARIES global_planner cell node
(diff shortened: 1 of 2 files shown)


LorenzMeier:
Thanks!

nicovanduijn:
Pretty sure it's gonna be required for #76, but until we decide to merge it this is fine

mhkabir:
@nicovanduijn What is it needed for specifically?

nicovanduijn:
I could be wrong about whether the dependency is actually needed or not. All I know is the mission interface uses the trajectory message  defined in ```mavros_msgs/Trajectory.h```  on the topics ```/mavros/trajectory/desired``` and ```/mavros/trajectory/generated```. Am I mixing something up?

mhkabir:
Yeah. That's `mavros_msgs`. `mav_msgs` is an old dependency from the `rotors_simulator` days.

nicovanduijn:
my bad, didn't read properly

PR #119, opened 2018-10-17
Add mutex before calling dynamic reconfigure

mrivi:
Fix for this warning
```[ WARN] [1539698071.890387845]: updateConfig() called on a dynamic_reconfigure::Server that provides its own mutex. This can lead to deadlocks if updateConfig() is called during an update. Providing a mutex to the constructor is highly recommended in this case. Please forward this message to the node author. ```

### DIFF
--- a/local_planner/src/nodes/local_planner_node.cpp
+++ b/local_planner/src/nodes/local_planner_node.cpp
@@ -5,15 +5,17 @@ LocalPlannerNode::LocalPlannerNode() {
   readParams();
 
   // Set up Dynamic Reconfigure Server
-  dynamic_reconfigure::Server<avoidance::LocalPlannerNodeConfig>::CallbackType
-      f;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: local_planner/src/nodes/local_planner_node.cpp [hunk 1]
```diff
@@ -5,16 +5,19 @@ LocalPlannerNode::LocalPlannerNode() {
...
   f = boost::bind(&LocalPlannerNode::dynamicReconfigureCallback, this, _1, _2);
-  server_.setCallback(f);
+  server_->setCallback(f);
 
   // disable memory if using more than one camera
   if (cameras_.size() > 1) {
-    rqt_param_config_.reproj_age_ = 0;
-    server_.updateConfig(rqt_param_config_);
-    dynamicReconfigureCallback(rqt_param_config_, 1);
+    if (config_mutex_.try_lock()) {
```
bkueng: Why not `config_mutex_.lock()`?

mrivi:
@bkueng fixed

Issue #118, opened 2018-10-16
Timeouts


mrivi:
I get timeouts messages `Pointcloud timeout , no cloud received on topic  (Hovering at current position) ` both in the simulation and on hardware.

I have already spoken offline with @baumanta. As she suspected, the timeout happens because the transformation between the camera frame and local origin is not available.

Opening a issue so that we track.

mrivi:
This is what I have discovered so far. The `canTransform` method gives the following error:
```camera 0 Lookup would require extrapolation into the future.  Requested time 68.824000000 but the latest data is at time 68.802956566, when looking up transform from frame [front_camera_link] to frame [local_origin]
camera 1 Lookup would require extrapolation into the future.  Requested time 68.824000000 but the latest data is at time 68.802956566, when looking up transform from frame [left_camera_link] to frame [local_origin]
camera 2 Lookup would require extrapolation into the future.  Requested time 68.824000000 but the latest data is at time 68.802956566, when looking up transform from frame [right_camera_link] to frame [local_origin]
```
The transformation is available only when the timestamp of the pointcloud matches the timestamp of the transformation.

Running the `tf_monitor` gives this result
```RESULTS: for local_origin to fcu
Chain is: local_origin -> fcu
Net delay     avg = 0.0640475: max = 0.0850794

Frames:
Frame: fcu published by unknown_publisher Average Delay: 0.0219803 Max Delay: 0.0655908

All Broadcasters:
Node: unknown_publisher 198.177 Hz, Average Delay: -0.00338286 Max Delay: 0.0655908
Node: unknown_publisher(static) 1e+08 Hz, Average Delay: 0 Max Delay: 0
```
As you can see in the frame view I have both the transformation from local_origin->fcu amd fcu->camera frame running at ~50Hz. While each point cloud is received at 20Hz.
[frames.pdf](https://github.com/PX4/avoidance/files/2496109/frames.pdf)

I went back to a older master (when the support for multiple pointcloud was not yet there) and it often happen that for one cycle the transformation is not available. Since it wasn't enough to show a timeout on the local_planner we never noticed.

mrivi:
fixed in #122

PR #122, opened 2018-10-23
Fix to avoid timeouts in the panner due to not receiving transforms

baumanta:
Before the transform at the exact time of the pointcloud was requested. As the realsense cameras and the pixhawk are not synchronized, this could cause timeouts. Now the closest transform in time is used, which got rid of the timeouts. This is obviously not as correct as using the exact transform, but it did not seem to have an influence on the performance in SITL
This would fix issue [118](https://github.com/PX4/avoidance/issues/118)


baumanta:
Here some results from running on the IF750A

The pointcloud come in with about 30 Hz:
   average rate: 30.045
   min: 0.012s max: 0.049s std dev: 0.00252s window: 1503

The transform if published by mavros with about 50Hz
RESULTS: for local_origin to fcu
Chain is: local_origin -> fcu
Net delay     avg = 0.00941085: max = 0.0371979

PR #130, opened 2018-10-31
local_planner_aeroD415: update to new pointcloud paramter name

mrivi:


baumanta:
Looks good to me. Do you still need that launch file for something? Didi you ever fly again with aero?

mrivi:
@baumanta no, but someone was using it and it was complaining on slack that it was not working.
We should either remove it or we need to keep maintaining it. I am fine with both options.

baumanta:
I'm gonna merge it for now. At some point we should think about cleaning up the launch files, maybe figure out a way in which we can have one base launch file and a parameter file, such that we do not need a new launch file for every platform

PR #131, opened 2018-11-01
Visualize drone in RViz

baumanta:
Making SITL great again. Instead of looking at the fcu frame to find the drone position, we now have a drone mesh in rviz.


baumanta:
Sadly the trees are still white. Seems to be an unresolved bug in ROS kinetic that Rviz does not render collada colors

PR #137, opened 2018-11-07
Modify FOV via a parameter.

baumanta:
As all hardware setups as well as SITL models have different fields of view, it is important to not have that hardcoded anymore. Especially, if the obstacle distance is sent to fcu, data is discarded if the FOV is set too small

### DIFF
--- a/local_planner/src/nodes/local_planner.h
+++ b/local_planner/src/nodes/local_planner.h
@@ -186,6 +186,8 @@ class LocalPlanner {
 
  public:
 
+  double H_FOV_ = 59.0;
+  double V_FOV_ = 46.0;
   Box histogram_box_;
   bool use_vel_setpoints_;
   bool currently_armed_ = false;
--- a/local_planner/src/nodes/local_planner.h
+++ b/local_planner/src/nodes/local_planner.h
@@ -186,8 +186,8 @@ class LocalPlanner {
 
  public:
 
-  double H_FOV_ = 59.0;
-  double V_FOV_ = 46.0;
+  double h_FOV_ = 59.0;
+  double v_FOV_ = 46.0;
   Box histogram_box_;
   bool use_vel_setpoints_;
   bool currently_armed_ = false;
--- a/local_planner/src/nodes/local_planner_node.cpp
+++ b/local_planner/src/nodes/local_planner_node.cpp
@@ -124,6 +124,13 @@ void LocalPlannerNode::readParams() {
   nh_.param<double>("min_speed_close_to_goal", new_params.min_speed_close_to_goal, 0.5);
 
   wp_generator_.param_ = new_params;
+
+  //set field of view
+  nh_.param<double>("horizontal_FOV", local_planner_.H_FOV_, 59.0);
+  nh_.param<double>("vertical_FOV", local_planner_.V_FOV_, 46.0);
+  local_planner_.H_FOV_ = local_planner_.H_FOV_;
+  local_planner_.V_FOV_ = local_planner_.V_FOV_;
+  wp_generator_.setFOV(local_planner_.H_FOV_, local_planner_.V_FOV_);
 }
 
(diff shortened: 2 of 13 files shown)


local_planner/src/nodes/local_planner.h
```diff
@@ -186,6 +186,8 @@ class LocalPlanner {
 
  public:
 
+  double H_FOV_ = 59.0;
```
mrivi: usually only constants/defines are capitalized

local_planner/src/nodes/local_planner_node.cpp
```diff
@@ -124,6 +124,13 @@ void LocalPlannerNode::readParams() {
   nh_.param<double>("min_speed_close_to_goal", new_params.min_speed_close_to_goal, 0.5);
 
   wp_generator_.param_ = new_params;
+
+  //set field of view
+  nh_.param<double>("horizontal_FOV", local_planner_.H_FOV_, 59.0);
+  nh_.param<double>("vertical_FOV", local_planner_.V_FOV_, 46.0);
+  local_planner_.H_FOV_ = local_planner_.H_FOV_;
```
mrivi: why is this needed?

local_planner/src/nodes/local_planner.h
baumanta: do you want me to decapitalize it? It was like that before, because it kinda is a constant that is only defined at the beginning but cannot make it const as I need to set it

local_planner/src/nodes/local_planner_node.cpp
baumanta: not needed

mhkabir:
Can't we compute the FOV directly from the camera calibration already available to us?

mrivi:
@mhkabir true, we have already everything available on the topic `*/camera_info`.

baumanta:
@mhkabir , @LorenzMeier , @mrivi  I agree, that would be a much cleaner solution. But the camera_info topic only contains  the image size. What we need is the pointcloud FOV, using USB2 that might be smaller than the image. I would have to verify that those numbers actually match

LorenzMeier:
It has the camera intrinsics as well, so FOV is there.

mrivi:
@baumanta from the image dimensions and the camera matrix you can calculate the field of view.

baumanta:
What I meant is that the USB 2 connection drops a part of the pointcloud FOV in some cases. Depending on resolution fps etc. But I think it is definitely doable. If the camera intrinsics do not fit the actual FOV we get, we could also infer the actual FOV from the received points

mhkabir:
> What I meant is that the USB 2 connection drops a part of the pointcloud FOV in some cases.

mrivi:
From `/camera/depth/camera_info` on a realsense D435
```
header: 
  seq: 1066
  stamp: 
    secs: 1541609101
    nsecs: 151765049
  frame_id: "right_camera_depth_optical_frame"
height: 270
width: 480
distortion_model: "plumb_bob"
D: [0.0, 0.0, 0.0, 0.0, 0.0]
K: [239.92958068847656, 0.0, 237.95184326171875, 0.0, 239.92958068847656, 129.16334533691406, 0.0, 0.0, 1.0]
R: [1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 1.0]
P: [239.92958068847656, 0.0, 237.95184326171875, 0.0, 0.0, 239.92958068847656, 129.16334533691406, 0.0, 0.0, 0.0, 1.0, 0.0]
binning_x: 0
binning_y: 0
roi: 
  x_offset: 0
  y_offset: 0
  height: 0
  width: 0
  do_rectify: False
```

Doing the math:
h_fov = 2 * atan(480 / (2 * 239.929)) = 90.01 deg
v_fov = 2 * atan(270 / (2 * 239.929)) = 58.72 deg

PR #139, opened 2018-11-15
Fov Refactor

mrivi:
This PR calculates the FOV from the camera_info topic instead of using a user supplied parameter.
Still missing
- account for overlap in the fov with multiple cameras
- account for different cameras in multi camera setup

PR is huge because of running clang format :(

### DIFF
--- a/local_planner/launch/local_planner_A700_1cam.launch
+++ b/local_planner/launch/local_planner_A700_1cam.launch
@@ -63,8 +63,6 @@
       <param name="goal_x_param" value="15" />
       <param name="goal_y_param" value="15"/>
       <param name="goal_z_param" value="4" />
-      <param name="horizontal_FOV" value="90" />
-      <param name="vertical_FOV" value="59" />
       <rosparam param="pointcloud_topics" subst_value="True">$(arg pointcloud_topics)</rosparam> 
     </node>
   
--- a/local_planner/launch/local_planner_A700_3cam.launch
+++ b/local_planner/launch/local_planner_A700_3cam.launch
@@ -97,8 +97,6 @@
       <param name="goal_x_param" value="15" />
       <param name="goal_y_param" value="15"/>
       <param name="goal_z_param" value="4" />
-      <param name="horizontal_FOV" value="270" />
-      <param name="vertical_FOV" value="59" />
       <rosparam param="pointcloud_topics" subst_value="True">$(arg pointcloud_topics)</rosparam> 
     </node>
     
--- a/local_planner/launch/local_planner_aero.launch
+++ b/local_planner/launch/local_planner_aero.launch
@@ -42,12 +42,13 @@
 
   <env name="ROSCONSOLE_CONFIG_FILE" value="$(find local_planner)/resource/custom_rosconsole.conf"/>
   <rosparam command="load" file="$(find local_planner)/cfg/params.yaml"/>
+  <arg name="pointcloud_topics" default="[/camera/depth/points]"/>
 
   <node name="local_planner_node" pkg="local_planner" type="local_planner_node" output="screen" >
     <param name="goal_x_param" value="15" />
     <param name="goal_y_param" value="15"/>
     <param name="goal_z_param" value="4" />
-    <param name="depth_points_topic" value="/camera/depth/points" />
+    <rosparam param="pointcloud_topics" subst_value="True">$(arg pointcloud_topics)</rosparam>
   </node>
 
 </launch>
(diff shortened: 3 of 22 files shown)


baumanta:
Did you check that the camera_info topic actually contains the correct FOV, even if the camera is attached over USB2?

mrivi:
yes, this are the result I got https://github.com/PX4/avoidance/pull/137#issuecomment-436694969

baumanta:
Ok, so it seems to match for our exact setup. But I think our pointcloud is not reduced in size because I set down the resolution and the rate. As soon as you set them up, I know that the pointcloud from the realsense gets smaller. But I was wondering whether this is also reflected in the camera_info. Because if it is not seen there this would lead to a rather dangerous corner case. If we assume the FOV is much larger than it actually is, we could run into stuff

mrivi:
I guess you have changed resolution and rate in the launch file. If so, it will be reflected in the camera info.

baumanta:
yes the resolution and rate will for sure be there. But as the USB2 connection does not allow for a pointcloud of that size/rate it will be smaller than demanded. And I'm not sure if the camera_info topic only reflects the demanded quantities or the actually executed ones. The camera info topic is the one for the depth image /camera/depth/camera_info, right? Maybe that only reflects the depth image size and not the actual size of the produced pointcloud. I'm not sure if in the case, where the pointcloud is reduced in size, also the depth image is reduced or not.

mrivi:
I think I am missing something. How is this different that setting 270deg and 59deg from a parameter? It's static and doesn't take into account a smaller than demanded pointcloud.

In any case, how do I reproduce the scenario when the realsense downsizes the pointcloud?

baumanta:
You are right, having the parameters static does not really help. I just thought, this way they are visible and everybody is responsible for checking whether they are right for their specific scenario. If it is automated, nobody knows what happens in the background and if the values are wrong, this will be very difficult to debug. 
But maybe I'm worrying too much about that corner case with the realsense. It's just that I know it is there and I assume many people will start using D435 over USB2 once they realize USB3 is jamming all their other stuff on the drone. 

You can reproduce it quite easily. If you set up the resolution in the rs_depthcloud.launch to the next higher supported value and maybe also the rate form 30 to 60 fps you will see that cloud suddenly is smaller.

baumanta:
I checked on my hardware setup. It seems the camera info matches the pointcloud size even when the FOV is reducted. 

Results:

resolution [480,270]
FOV measured from received points [89, 58]
FOV calculated form camera info [89.8, 58.5]
camera info:
header: 
  seq: 2774
  stamp: 
    secs: 1542615853
    nsecs: 963099917
  frame_id: "front_camera_depth_optical_frame"
height: 270
width: 480
distortion_model: "plumb_bob"
D: [0.0, 0.0, 0.0, 0.0, 0.0]
K: [241.01246643066406, 0.0, 239.676513671875, 0.0, 241.01246643066406, 136.3378448486328, 0.0, 0.0, 1.0]
R: [1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 1.0]
P: [241.01246643066406, 0.0, 239.676513671875, 0.0, 0.0, 241.01246643066406, 136.3378448486328, 0.0, 0.0, 0.0, 1.0, 0.0]
binning_x: 0
binning_y: 0
roi: 
  x_offset: 0
  y_offset: 0
  height: 0
  width: 0
  do_rectify: False


resolution [640,480]
FOV measured from received points [78, 64]
FOV calculated form camera info [79.373979, 63.7941505]
camera info:
header: 
  seq: 3103
  stamp: 
    secs: 1542616174
    nsecs:  36993532
  frame_id: "front_camera_depth_optical_frame"
height: 480
width: 640
distortion_model: "plumb_bob"
D: [0.0, 0.0, 0.0, 0.0, 0.0]
K: [385.61993408203125, 0.0, 319.482421875, 0.0, 385.61993408203125, 242.14056396484375, 0.0, 0.0, 1.0]
R: [1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 1.0]
P: [385.61993408203125, 0.0, 319.482421875, 0.0, 0.0, 385.61993408203125, 242.14056396484375, 0.0, 0.0, 0.0, 1.0, 0.0]
binning_x: 0
binning_y: 0
roi: 
  x_offset: 0
  y_offset: 0
  height: 0
  width: 0
  do_rectify: False

baumanta:
Do you want me to merge that?

mrivi:
@baumanta wait, I need to modify the callback. Right now I am unsubscribing from the topic after setting the FOV for the first. I need to keep the subscription such that we can catch the case that the pointcloud is downsized

baumanta:
@mrivi not sure thta is necessary. in my test the first camera info I got was already correct. It seems that logic for reduction is done before anything is published from what I can see

mrivi:
@baumanta ah sorry, I misunderstood. It's good to go then

PR #140, opened 2018-11-15
[RFC] add generic launch file for local planner

mrivi:
@baumanta @hamishwillee - Since we started talking about the documentation, I got the idea to create a generic launch file which contains also comments where the user needs to modify things specific to his/her setup.
We could then remove all the HW specific launch files a part maybe for the Auterion platform. I think we have too many launch files and it hard to keep them in sink with the code base especially since we don't have any CI.


mrivi:
yes, it's meant as an example. You're right, the name choice is a bit unfortunate.
I would then remove all HW specific launch files a part from this one and the A700_3cam.

I am not sure if it's too difficult to spot the comments where the user has to actually edit something. That's why with the camera options I tried that `*****` separation.

baumanta:
I agree. But I would request, that the local_planner_sitl_3cam.launch stays as well. This is actually the file I use most often and it would be a bit of a pain to have that locally. Also I think it would be nice to have one hardware and one sitl launch file as working examples for the user. What do you think?

mrivi:
yes, this is only for hw setups. I haven't removed any of the simulation files.

We can clean up the the simulation launch files in a separate PR.

hamishwillee:
@mrivi So the idea is that if you want to use this system you copy the file, edit it according to the comments, and then run your own new launch file?

Fewer files to maintain is good, but then so is "click this file and go". Anywhere you're asking people to edit things you're asking for trouble :-)
In other systems I tend to lean towards command line options with sensible defaults over editable files - but I don't know what ROS allows.

It seems sensible and I assume this is something you would like to use - but not being a ROS developer it is not clear to me whether an audience would like this approach or not. My goto guy on this sort of question is @TSC21 :-)

local_planner/launch/example.launch
```diff
@@ -1,36 +1,28 @@
...
-    <arg name="mavros_transformation" default="0" />
-    
     <param name="use_sim_time" value="false" />
 
-    <node pkg="tf" type="static_transform_publisher" name="tf_90_deg"
-          args="0 0 0 $(arg mavros_transformation) 0 0 world local_origin 10"/>
+    <!-- Transformation between the fcu and camera frame. It needs to be changed
+    according to the specific setup. Please measure the dispacement between the
+    camera and the flight controller in ENU body frame and update the x, y, z,
+    yaw, pitch, and roll in meters-->
```
hamishwillee: So a question I might ask here is which are the x, y, z etc params. (ie you need to confirm that they are the `args=` and ordered as you say.

local_planner/launch/example.launch
```diff
@@ -1,36 +1,28 @@
...
           args="0.30 0.32 -0.11 0 0 0 fcu front_camera_link 10"/>
-                
+
     <arg name="headless" default="false"/>
     <arg name="ns" default="/"/>
     <arg name="build" default="posix_sitl_default"/>
-    <arg name="fcu_url" default="udp://:14540@localhost:14557"/>
-    <arg name="gcs_url" default="" />   <!-- GCS link is provided by SITL -->
+
+    <!-- Change the serial or udp port to match your setup-->
```
hamishwillee: And if you have something that should be changed I'd separate it from everything by a line below too. Also be very clear about the field being changed.
Maybe like:
```
<!-- Change fcu_url arg default attribute to the port used in your setup (serial or udp) -->
    <arg name="fcu_url" default="/dev/ttyACM0:57600"/>

    <arg name="gcs_url" default="" />
```

local_planner/launch/example.launch
```diff
@@ -39,36 +31,61 @@
         </include>
     </group>
 
-    <!-- Launch cameras -->  
+    <!-- Examples of how to launch cameras. Select the one appropariate. -->
```
hamishwillee: Typo appropariate

mrivi:
@hamishwillee I am trying to solve a couple of problems:
- people need to edit the launch file for some parameters to have the system working properly
- we have too many launch files and it has already happen that poeple used one out-of-date and had issues

Yes, all arguments can be edited by command line. For example:
`roslaunch local_planner local_planner_*.launch fcu_url:="/dev/ttyTHS1:57600" `

My idea was to leave the local_planner_A700_3cam.launch which is ready to use (need to modify transformation, fcu_url and camera serial number ) with a D435 realsense camera (which is the most widespread camera for what I have seen so far)

TSC21:
@hamishwillee as Martina said, the parameters of ROS nodes can be parsed as arguments on the command line. I always try to make things flexible and configurable as much as possible, only limiting the configs according to the end user/developer. For an open solution, I think @mrivi's approach is the correct one.

baumanta:
@mrivi @TSC21 I agree that this would be the cleanest way to do it, as it avoids launch files getting out of date. I just see a few disadvantages for our tests. Because we are using resin OS (docker containers) for testing, the repo (with a certain branch) is included in the base image. Every local change is deleted whenever we restart the docker. Therefore, it would be much more convenient if the launch file we need were inside that repo. If they weer not, we would always need a custom branch with the launch files or some additional scripts outside the repo in a permanent space withthe commands using the correct parameters. It would just be a bit more cumbersome than having A700_1cam and A700_cam launch files in the repo.

baumanta:
@TSC21 we could do that. It's just that the parameters include a lot of long numbers like udp ports and realsense serial numbers. Therefore, typing them would be cumbersome. So we would need an separate file in the permanent space with the launch command and all the parameters. We could do that, but we are trying to limit those files to a minimum therefore, it would just be more convenient the other way.

 Also a launch command with all the custom parameter would look quite ugly as you need to specify:
-udp port
-camera serial numbers (in our case three or four of them)
-pointcloud topic names
-all the static transforms belonging to the cameras

Only if all of that are parameters we cold have one launch file for everyone

baumanta:
I was thinking about xml parameter files. Such we could have a lot of them for every drone and configuration and everyone can have its own. Those would contain the mentioned numbers. We would only have one launch file with the xml file as an argument. I have no idea if that works, but it might be a cleaner more practical solution.

TSC21:
@baumanta for that case, better to use YAML format: http://wiki.ros.org/roslaunch/XML/rosparam

TSC21:
> @mrivi @TSC21 yaml sounds nice. What do you think, would that be an option? Having some yaml files which contain the custom numbers and only one (or maybe two, I don't know if we can also specify the number of cameras which should be launched in the yaml) launch file which needs maintaining.

That seems reasonable.

hamishwillee:
@mrivi et al. Thanks for clarifications - it sounds like you're all in agreement, so I'm good with this. 

> My idea was to leave the local_planner_A700_3cam.launch which is ready to use (need to modify transformation, fcu_url and camera serial number ) with a D435 realsense camera (which is the most widespread camera for what I have seen so far)

Possibly a dumb thought, but will other's have the same problems as @baumanta for the text? Would it be better/possible to supply a yaml file for local_planner_A700_3cam.launch + D435 realsense camera instead of the modified file and suggest people just modify the yaml for their version?

baumanta:
Yes @hamishwillee, I think that is exactly the idea. We will have one launch file only and a lot of yaml files, which contain all those serial numbers and transforms. That way we need to maintain only the one launch file and can provide nice examples for yaml files with comments on what belongs in there. Maybe a second launch file is required for a different number of cameras, as I'm not sure we can handle different  numbers of cameras from the same launch file. But we will have to try and see.

mrivi:
@baumanta @TSC21 It seems that it isn't possible to pass parameters from a yaml file to a launch file without some hacks. See https://answers.ros.org/question/213145/use-rosparam-get-in-launch-file/

mrivi:
@TSC21 the planner parameters are already handled by the rosparam server. 

I think in this refactor we are only talking about these specific arguments:
- static transformation parameters
- realsense serial number
- fcu_url

These parameters need to be edited either in the launch file or by command line

TSC21:
Ok that's a different story. It can make sense then to use the workaround of the link you shared.

PR #143, opened 2018-11-21
Cleanup local planner

jkflying:
Fix a number of warnings found with GCC -Wall and -Wextra. Do a tiny bit of refactoring based on some of the things that were brought up.

### DIFF
--- a/local_planner/src/nodes/common.h
+++ b/local_planner/src/nodes/common.h
@@ -18,38 +18,41 @@
 #include <pcl_conversions/pcl_conversions.h>
 
 float distance2DPolar(int e1, int z1, int e2, int z2);
-float computeL2Dist(geometry_msgs::Point position,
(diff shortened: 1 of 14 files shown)


local_planner/src/nodes/common.h
```diff
@@ -31,8 +31,11 @@ bool hasSameYawAndAltitude(geometry_msgs::PoseStamped old_wp,
                            double old_yaw);
 double elevationIndexToAngle(int e, double res);
 double azimuthIndexToAngle(int z, double res);
-int azimuthAnglefromCartesian(double x, double y, double z,
-                              geometry_msgs::Point pos);
+int azimuthAnglefromCartesian(geometry_msgs::Point position,
```
mrivi: these could be `const geometry_msgs::Point &position`

jkflying: I've done this now for all of the larger types in common

PR #148, opened 2018-11-27
Local planner smoke tests

jkflying:
Do some minor cleanup with references, argument types etc. 

More critically, add 4 high-level tests with the local planner showing the basic usage in different cases.

If you have any questions or see missing setup for the tests as written please please please comment.

### DIFF
--- a/local_planner/test/test_local_planner.cpp
+++ b/local_planner/test/test_local_planner.cpp
@@ -1,0 +1,49 @@
+#include <gtest/gtest.h>
+
+#include "../src/nodes/local_planner.h"
+
+// Stateless tests:
+// Create some hardcoded scan data of obstacles in different positions
+// For each one check that the planner response is correct
+
+class LocalPlannerTests : public ::testing::Test {
+ public:
+  LocalPlanner planner;
+
+  void SetUp() override {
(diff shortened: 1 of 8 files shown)


local_planner/test/test_local_planner.cpp
```diff
@@ -0,0 +1,216 @@
...
+  void TearDown() override {}
+};
+
+TEST_F(LocalPlannerTests, no_obstacles) {
+  // GIVEN: a local planner, a scan with no obstacles, pose and goal
+  planner.complete_cloud_.emplace_back();
+
+  // WHEN: we run the local planner
+  planner.runPlanner();
+  planner.runPlanner();
```
mrivi: why is it necessary to run the planner two times?

jkflying: `runPlanner()` only checks that it is at the correct altitude (`reach_altitude_`) at the end of the `runPlanner()`, so the first time it is run it doesn't actually do anything. I think there is an `else` which can probably be removed in `determineStrategy()` line 183 and instead checking `reach_altitude_` again, but I want a bit more testing before we start changing the logic of the planner.

PR #150, opened 2018-11-30
Status message

baumanta:
Make the local planner output a status message. This message is converted in mavros in a HEARTBEAT. soon there will be a PR merged on the Firmware side, which will output warnings over QGC when the avoidance system does not respond or reports timeouts

### DIFF
--- a/local_planner/src/nodes/local_planner_node.cpp
+++ b/local_planner/src/nodes/local_planner_node.cpp
@@ -940,7 +940,7 @@ int main(int argc, char** argv) {
   bool landing = false;
   avoidanceOutput planner_output;
   bool startup = true;
-  Node.status_msg_.state = 1; //COMPANION_PROCESS_STATE_STARTING
+  Node.status_msg_.state = (int)MAV_STATE::MAV_STATE_BOOT;
 
   std::thread worker(&LocalPlannerNode::threadFunction, &Node);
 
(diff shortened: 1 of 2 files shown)


baumanta:
https://github.com/PX4/Firmware/pull/10933

local_planner/src/nodes/local_planner_node.cpp
```diff
@@ -938,6 +940,7 @@ int main(int argc, char** argv) {
   bool landing = false;
   avoidanceOutput planner_output;
   bool startup = true;
+  Node.status_msg_.state = 1; //COMPANION_PROCESS_STATE_STARTING
```
jkflying: As soon as these are updated to the real enums I am happy with this.

baumanta: you are right this is ugly. The enum is already fixed and we know what it looks like. But it is defined in mavlink, therefore I now defined the same enum also in the local_planner such that we can use the same terms

local_planner/src/nodes/local_planner_node.h
```diff
@@ -47,6 +47,18 @@ struct cameraData {
   bool received_;
 };
 
+enum class MAV_STATE {
+  MAV_STATE_UNINIT,
```
jkflying: Ok, if these need to match an external number, I would initialize it with 0, or  even number them all. Also just add a little comment above the enum declaration here to say where this comes from.

PR #151, opened 2018-12-03
More local_planner cleanup

jkflying:
A number of small cleanups, which will make future testing and the migration to Eigen types easier.

This PR may be easier to review commit-by-commit, if the straight diff is a bit overwhelming.


local_planner/src/nodes/star_planner.cpp
```diff
@@ -1,23 +1,23 @@
...
+namespace avoidance {
+
+StarPlanner::StarPlanner() : tree_age_(0) {}
 
 StarPlanner::~StarPlanner() {}
 
 // set parameters changed by dynamic rconfigure
 void StarPlanner::dynamicReconfigureSetStarParams(
-    avoidance::LocalPlannerNodeConfig& config, uint32_t level) {
+    const avoidance::LocalPlannerNodeConfig& config, uint32_t level) {
```
nicovanduijn: Do we have a convention for the spacing? You seem to have changed the "&" to prepend the name instead of directly following the type, but it's not consistent across the code base. I don't care which convention we use, but I would prefer consistency

local_planner/src/nodes/star_planner.cpp
```diff
@@ -1,23 +1,23 @@
...
   childs_per_node_ = config.childs_per_node_;
   n_expanded_nodes_ = config.n_expanded_nodes_;
   tree_node_distance_ = config.tree_node_distance_;
   tree_discount_factor_ = config.tree_discount_factor_;
 }
 
-void StarPlanner::setParams(double min_cloud_size,
-                            double min_dist_backoff,
+void StarPlanner::setParams(double min_cloud_size, double min_dist_backoff,
```
nicovanduijn: I guess these doubles could be passed by reference as well

local_planner/src/nodes/star_planner.cpp
mrivi: we have a [script](https://github.com/PX4/avoidance/blob/master/tools/fix_style.sh) to style with clang-format. To run it: `$ ./tools/fix_style.sh local_planner`.  We would even have automatic build and style checks with [travis](https://github.com/PX4/avoidance/blob/master/.travis.yml#L72) but it isn't active

local_planner/src/nodes/local_planner.cpp
```diff
@@ -74,15 +77,21 @@ void LocalPlanner::dynamicReconfigureSetParams(
 }
 
 void LocalPlanner::setVelocity() {
-  velocity_mod_ = Eigen::Vector3f(curr_vel_.twist.linear.x,
-    curr_vel_.twist.linear.y, curr_vel_.twist.linear.z).norm();
+  velocity_mod_ =
+      Eigen::Vector3f(curr_vel_.twist.linear.x, curr_vel_.twist.linear.y,
+                      curr_vel_.twist.linear.z)
+          .norm();
```
baumanta: This does look a bit weird from a formatting perspective. Is it supposed to be this way?

local_planner/src/nodes/star_planner.cpp
baumanta: I would not pass doubles by reference I thought it only makes sense really for bigger entities, no?

jkflying: Yeah, pass by reference is just extra complications for small types. No performance differences (in fact, possibly worse), and it makes it harder to read.

local_planner/src/nodes/local_planner.cpp
jkflying: This is just clang-format... We could manually do formatting, but once this is all Eigen vectors this will be a simple `velocity_mod_ = cur_vel.twist.linear.norm();` anyway.

baumanta: Ok, sure, that's what I suspected. I just find this particular line pretty hard to read in clang. But you are right with Eigen it is gonna change anyway

jkflying: Ok, fixed :)

local_planner/src/nodes/star_planner.cpp
jkflying: Yes, this isn't my 'ideal' formatting style in terms of readability, but at least it is consistent and automated.

PR #153, opened 2018-12-05
Eigen conversions part 1

jkflying:
This converts a bunch of the types and calculations to use Eigen instead of a mix of pcl::PointXYZ, mavros_msg::Point and tf::Vector3. Would rather get this partial work merged to avoid merge conflicts while the rest is being done.

Please double-check the places where calculations were simplified, the arguments and types are mostly checked by the compiler.

Code formatted and tests pass :heavy_check_mark:

### DIFF
--- a/local_planner/src/nodes/common.cpp
+++ b/local_planner/src/nodes/common.cpp
@@ -61,29 +61,29 @@ double azimuthIndexToAngle(int z, double res) {
   return z * res + res / 2 - 180;
 }
 
-int azimuthAnglefromCartesian(const geometry_msgs::Point& position,
-                              const geometry_msgs::Point& origin) {
-  return azimuthAnglefromCartesian(position.x, position.y, origin);
+int azimuthAnglefromCartesian(const Eigen::Vector3f& position,
+                              const Eigen::Vector3f& origin) {
+  return azimuthAnglefromCartesian(position.x(), position.y(), origin);
 }
 
-int azimuthAnglefromCartesian(double x, double y,
-                              const geometry_msgs::Point& pos) {
-  return floor(atan2(x - pos.x, y - pos.y) * 180.0 / M_PI);  //(-180. +180]
+int azimuthAnglefromCartesian(double x, double y, const Eigen::Vector3f& pos) {
+  return floor(atan2(x - pos.x(), y - pos.y()) * 180.0 / M_PI);  //(-180. +180]
 }
 
 int elevationAnglefromCartesian(double x, double y, double z,
-                                const geometry_msgs::Point& pos) {
-  double den = sqrt((x - pos.x) * (x - pos.x) + (y - pos.y) * (y - pos.y));
+                                const Eigen::Vector3f& pos) {
+  double den =
+      sqrt((x - pos.x()) * (x - pos.x()) + (y - pos.y()) * (y - pos.y()));
   if (den == 0) {
     return 0;
   } else {
-    return floor(atan((z - pos.z) / den) * 180.0 / M_PI);  //(-90.+90)
+    return floor(atan((z - pos.z()) / den) * 180.0 / M_PI);  //(-90.+90)
   }
 }
 
-int elevationAnglefromCartesian(const geometry_msgs::Point& pos,
-                                const geometry_msgs::Point& origin) {
-  return elevationAnglefromCartesian(pos.x, pos.y, pos.z, origin);
+int elevationAnglefromCartesian(const Eigen::Vector3f& pos,
+                                const Eigen::Vector3f& origin) {
+  return elevationAnglefromCartesian(pos.x(), pos.y(), pos.z(), origin);
 }
 
 int elevationAngletoIndex(int e, int res) {  //[-90,90]
--- a/local_planner/src/nodes/common.cpp
+++ b/local_planner/src/nodes/common.cpp
@@ -207,16 +207,51 @@ double getAngularVelocity(double desired_yaw, double curr_yaw) {
   return 0.5 * vel;
 }
 
+Eigen::Vector3f toEigen(const geometry_msgs::Point& p) {
+  Eigen::Vector3f ev3(p.x, p.y, p.z);
+  return ev3;
+}
+Eigen::Vector3f toEigen(const pcl::PointXYZ& p) {
+  Eigen::Vector3f ev3(p.x, p.y, p.z);
+  return ev3;
+}
+Eigen::Vector3f toEigen(const geometry_msgs::Vector3& p) {
+  Eigen::Vector3f ev3(p.x, p.y, p.z);
+  return ev3;
+}
+
+geometry_msgs::Point toPoint(const Eigen::Vector3f& ev3) {
+  geometry_msgs::Point gmp;
+  gmp.x = ev3.x();
+  gmp.y = ev3.y();
+  gmp.z = ev3.z();
+  return gmp;
+}
+pcl::PointXYZ toXYZ(const Eigen::Vector3f& ev3) {
+  pcl::PointXYZ gmp;
+  gmp.x = ev3.x();
+  gmp.y = ev3.y();
+  gmp.z = ev3.z();
+  return gmp;
+}
+geometry_msgs::Vector3 toVector3(const Eigen::Vector3f& ev3) {
+  geometry_msgs::Vector3 gmp;
+  gmp.x = ev3.x();
+  gmp.y = ev3.y();
+  gmp.z = ev3.z();
+  return gmp;
+}
+
 Eigen::Vector3f convert(const geometry_msgs::Point& p) {
   Eigen::Vector3f ev3(p.x, p.y, p.z);
   return ev3;
 }
 
-geometry_msgs::Point convert(const Eigen::Vector3f& p) {
+geometry_msgs::Point convert(const Eigen::Vector3f& ev3) {
   geometry_msgs::Point gmp;
-  gmp.x = p.x();
-  gmp.y = p.y();
-  gmp.z = p.z();
+  gmp.x = ev3.x();
+  gmp.y = ev3.y();
+  gmp.z = ev3.z();
   return gmp;
 }
 }
--- a/local_planner/src/nodes/common.cpp
+++ b/local_planner/src/nodes/common.cpp
@@ -13,26 +13,14 @@ float distance2DPolar(int e1, int z1, int e2, int z2) {
   return sqrt(pow((e1 - e2), 2) + pow((z1 - z2), 2));
 }
 
-float computeL2Dist(const geometry_msgs::Point& position,
-                    const pcl::PointXYZ& xyz) {
-  return sqrt(pow(position.x - xyz.x, 2) + pow(position.y - xyz.y, 2) +
-              pow(position.z - xyz.z, 2));
-}
-
-float distance3DCartesian(const geometry_msgs::Point& a,
-                          const geometry_msgs::Point& b) {
-  return sqrt((a.x - b.x) * (a.x - b.x) + (a.y - b.y) * (a.y - b.y) +
-              (a.z - b.z) * (a.z - b.z));
-}
-
 // transform polar coordinates into Cartesian coordinates
-geometry_msgs::Point fromPolarToCartesian(int e, int z, double radius,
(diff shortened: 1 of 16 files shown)


local_planner/src/nodes/common.cpp
```diff
@@ -13,26 +13,14 @@ float distance2DPolar(int e1, int z1, int e2, int z2) {
...
-              pow(position.z - xyz.z, 2));
-}
-
-float distance3DCartesian(const geometry_msgs::Point& a,
-                          const geometry_msgs::Point& b) {
-  return sqrt((a.x - b.x) * (a.x - b.x) + (a.y - b.y) * (a.y - b.y) +
-              (a.z - b.z) * (a.z - b.z));
-}
-
 // transform polar coordinates into Cartesian coordinates
```
nicovanduijn: I think we should make a big note here on the units. The fact that `e` and `z` are integers (which I find questionable in the first place in this context), and represent elevation and orientation in degrees should be documented somewhere, no?

local_planner/src/nodes/common.cpp
```diff
@@ -13,26 +13,14 @@ float distance2DPolar(int e1, int z1, int e2, int z2) {
...
-  p.x =
+Eigen::Vector3f fromPolarToCartesian(int e, int z, double radius,
+                                     const geometry_msgs::Point& pos) {
+  Eigen::Vector3f p;
+  p.x() =
       pos.x + radius * cos(e * (M_PI / 180)) * sin(z * (M_PI / 180));  // round
-  p.y = pos.y + radius * cos(e * (M_PI / 180)) * cos(z * (M_PI / 180));
-  p.z = pos.z + radius * sin(e * (M_PI / 180));
+  p.y() = pos.y + radius * cos(e * (M_PI / 180)) * cos(z * (M_PI / 180));
+  p.z() = pos.z + radius * sin(e * (M_PI / 180));
```
nicovanduijn: Lots of implicit type conversions here. Mixing int, float and double... Since you seem to have chosen Vector3f, maybe single-precision floating points all around?

local_planner/src/nodes/common.cpp
```diff
@@ -75,29 +61,29 @@ double azimuthIndexToAngle(int z, double res) {
...
-  return azimuthAnglefromCartesian(position.x, position.y, origin);
+int azimuthAnglefromCartesian(const Eigen::Vector3f& position,
+                              const Eigen::Vector3f& origin) {
+  return azimuthAnglefromCartesian(position.x(), position.y(), origin);
 }
 
-int azimuthAnglefromCartesian(double x, double y,
-                              const geometry_msgs::Point& pos) {
-  return floor(atan2(x - pos.x, y - pos.y) * 180.0 / M_PI);  //(-180. +180]
+int azimuthAnglefromCartesian(double x, double y, const Eigen::Vector3f& pos) {
```
nicovanduijn: Again, a note that this is in degrees would be nice. Also I'm still confused why this needs to be an integer

local_planner/src/nodes/common.cpp
```diff
@@ -207,16 +193,51 @@ double getAngularVelocity(double desired_yaw, double curr_yaw) {
...
+}
+geometry_msgs::Vector3 toVector3(const Eigen::Vector3f& ev3) {
+  geometry_msgs::Vector3 gmp;
+  gmp.x = ev3.x();
+  gmp.y = ev3.y();
+  gmp.z = ev3.z();
+  return gmp;
+}
+
 Eigen::Vector3f convert(const geometry_msgs::Point& p) {
```
nicovanduijn: Do we still need this? What's the difference to `Eigen::Vector3f toEigen(const geometry_msgs::Point& p) ` ?

local_planner/src/nodes/common.cpp
```diff
@@ -207,16 +193,51 @@ double getAngularVelocity(double desired_yaw, double curr_yaw) {
...
   Eigen::Vector3f ev3(p.x, p.y, p.z);
   return ev3;
 }
 
-geometry_msgs::Point convert(const Eigen::Vector3f& p) {
+geometry_msgs::Point convert(const Eigen::Vector3f& ev3) {
```
nicovanduijn: Also, I believe this is legacy since the addition of `toPoint()` ?

local_planner/src/nodes/local_planner.cpp
```diff
@@ -205,10 +206,10 @@ void LocalPlanner::determineStrategy() {
         reach_altitude_ && use_back_off_) {
       if (!back_off_) {
         back_off_point_ = closest_point_;
-        back_off_start_point_ = pose_.pose.position;
+        back_off_start_point_ = toEigen(pose_.pose.position);
```
nicovanduijn: Never been a fan of the `pose_.pose`-business. IMHO `pose_` should be more descriptive. Pose of what? The goal? The drone? Donald Trump? We don't know...

local_planner/src/nodes/planner_functions.cpp
```diff
@@ -105,13 +90,9 @@ void filterPointCloud(
   }
 
   if (counter_sphere > 0) {
-    temp_sphere_center.x = temp_sphere_center.x / counter_sphere;
-    temp_sphere_center.y = temp_sphere_center.y / counter_sphere;
-    temp_sphere_center.z = temp_sphere_center.z / counter_sphere;
+    temp_sphere_center /= counter_sphere;
```
nicovanduijn: I'm not familiar enough with the Eigen library, but this looks suspicious. Do you indeed have to call the `array()` method for the += operator, but not for the /= ? Why?

local_planner/src/nodes/planner_functions.cpp
```diff
@@ -175,11 +156,9 @@ void propagateHistogram(Histogram &polar_histogram_est,
                         geometry_msgs::PoseStamped position) {
```
nicovanduijn: Sorry for all the comments that are somewhat unrelated to your changes, but I just noticed this one. A `PoseStamped position` leads to the super beautiful construct `position.pose.position` down the road

local_planner/src/nodes/waypoint_generator.cpp
```diff
@@ -200,30 +184,25 @@ bool WaypointGenerator::withinGoalRadius() {
...
 void WaypointGenerator::reachGoalAltitudeFirst() {
-  output_.goto_position.x = planner_info_.offboard_pose.pose.position.x;
-  output_.goto_position.y = planner_info_.offboard_pose.pose.position.y;
+  output_.goto_position = planner_info_.offboard_pose.pose.position;
   output_.goto_position.z = pose_.pose.position.z + 0.5;
 
   // if goal lies directly overhead, do not yaw
-  double x_diff = std::abs(goal_.x - pose_.pose.position.x);
-  double y_diff = std::abs(goal_.y - pose_.pose.position.y);
+  Eigen::Vector3f diff = goal_ - toEigen(pose_.pose.position);
```
nicovanduijn: Where did the absolute value go? May lead to faulty behavior in line 193

local_planner/src/nodes/waypoint_generator.cpp
```diff
@@ -200,30 +184,25 @@ bool WaypointGenerator::withinGoalRadius() {
...
-  pose_to_wp.x *= planner_info_.min_speed;
-  pose_to_wp.y *= planner_info_.min_speed;
-  pose_to_wp.z *= planner_info_.min_speed;
-  output_.goto_position.x = pose_.pose.position.x + pose_to_wp.x;
-  output_.goto_position.y = pose_.pose.position.y + pose_to_wp.y;
-  output_.goto_position.z = pose_.pose.position.z + pose_to_wp.z;
+  Eigen::Vector3f pose_to_wp =
+      toEigen(output_.goto_position) - toEigen(pose_.pose.position);
+
+  if (pose_to_wp.norm() > 0.01) pose_to_wp.normalize();
```
nicovanduijn: What's our convention for one-line if statements? This same logic was used with different syntax in line 327

jkflying:
> All of these conversions are provided by ROS already.
> Please use https://github.com/ros/geometry `eigen_conversions` and `tf_conversions` as needed.

Thanks for the pointer. However, those APIs use a two argument with references style, which is much less readable when doing inline conversions of types. This might have made sense 10 years ago when RVO wasn't widely available in compilers, but today it is a headache. For example, this one-liner:
```
double dist = (toEigen(pose_.pose.position) - back_off_point_).norm();
```
turns into
```
Eigen::Vector3f eigenPosition;
tf::pointMsgToEigen(pose_.pose.position, eigenPosition);
double dist = (eigenPosition - back_off_point_).norm();
```
In terms of engineering tradeoffs I prefer the duplication of a few 2 - 5 line helper functions in order to maintain readability in core algorithm areas, where complicated code easily leads to bugs and makes visibility of the high level algorithm opaque.

If you think there is the possibility of bugs in these helpers I'm happy to wrap the `tf` versions, however it seems a bit overkill for simple assignment of xyz from one type to another.

local_planner/src/nodes/common.cpp
jkflying: Agreed, after the Eigen conversion everything should be floats.

local_planner/src/nodes/common.cpp
jkflying: Documentation would be much appreciated!

local_planner/src/nodes/common.cpp
jkflying: Good catch!

local_planner/src/nodes/waypoint_generator.cpp
jkflying: I prefer a single line with no `{}` if it all fits on a single line. But to be honest, either way doesn't bother me.

local_planner/src/nodes/local_planner.cpp
jkflying: I'll keep in mind a more descriptive name for it for the next PR, which will have to change the `pose_` things considerably.

local_planner/src/nodes/common.cpp
jkflying: Deleted now!

local_planner/src/nodes/common.cpp
jkflying: Yep, I want `float`s for everything here. `double`s are overkill (we aren't doing space probe navigation or mm-precision world-scale photogrammetry), and `int`s lead to strange behavior with rounding.

mrivi:
@jkflying should this PR be flight tested before merging? I feel this and the previous PR were quite substantial and it would be nice to test them without new features on top.

jkflying:
> @jkflying should this PR be flight tested before merging? I feel this and the previous PR were quite substantial and it would be nice to test them without new features on top.

Agreed, I'll hold off on merging until we've done some flight testing.

ploverpang:
I test master branch in SITL, both offboard and mission mode doesn't work as before. The drone is twitching and no longer flying normally. The speed always near 0.1m/s.

nicovanduijn:
I have to agree with @ploverpang.
Two logs from SITL:
-  pre-merge of this PR: (https://dev.auterion.com/plot_app?log=7be15f4d-6839-4b08-a09d-483a1e43d4da)
bokeh_plot 3
- After the merge : (https://dev.auterion.com/plot_app?log=a17df400-84be-4934-8291-b8a12353aae8)
bokeh_plot 4

baumanta:
I found the issue: It seems that (for some reason we still have to investigate) the planner runs slower now. Therefore, it triggers a timeout and goes to hover mode until it catches up again. So you can get it running by either building in Release mode (on my laptop that is enough), or by setting the parameter *pointcloud_timeout_hover_*  somewhat higher. If you are only using SITL this parameter is not crucial, you can use whatever works for you, if you want to run on a drone you would have to tune it for the setup. You can set this value over rqt_reconfigure or by modifying the default in local_planner.cfg

mhkabir:
I think that building Release by default would be a sensible option?

jkflying:
Defaulting to Debug is a `catkin` setting unfortunately - there are instructions on getting the Release builds to work in the Readme.md. We could maybe print some extra warnings on startup if `NDEBUG` isn't defined to help with this in the future.

It might also help to print separate warnings if the point cloud is slow vs. if the planner is slow to help debug situations like this.

PR #155, opened 2018-12-06
Change 1Cam launch file SN

nicovanduijn:
Mostly for convenience with testing on the UP board, this commit changes the S/N of the RealSense I am using. Also it's confusing to have three S/N in the launch file for just 1 RealSense.


nicovanduijn:
Turns out Tanja is still using the old file, so I'll do it differently

PR #156, opened 2018-12-06
Histogram image

baumanta:
Convert histogram into image format and pulish as an image. This provides a low-weight logging option

### DIFF
--- a/local_planner/src/nodes/local_planner.cpp
+++ b/local_planner/src/nodes/local_planner.cpp
@@ -167,6 +167,33 @@ void LocalPlanner::create2DObstacleRepresentation(const bool send_to_fcu) {
     updateObstacleDistanceMsg(to_fcu_histogram_);
   }
   polar_histogram_ = new_histogram;
+
+  //generate histogram image for logging
+  histogram_image_ = generateHistogramImage(polar_histogram_);
+}
+
+sensor_msgs::Image LocalPlanner::generateHistogramImage(Histogram& histogram) {
+
+	sensor_msgs::Image image;
+	double sensor_max_dist = 20.0;
+	image.header.stamp = ros::Time::now();
+	image.height = GRID_LENGTH_E;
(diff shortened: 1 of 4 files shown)


local_planner/src/nodes/local_planner.cpp
```diff
@@ -167,6 +167,33 @@ void LocalPlanner::create2DObstacleRepresentation(const bool send_to_fcu) {
...
+	image.encoding = sensor_msgs::image_encodings::MONO8;
+	image.is_bigendian = 0;
+	image.step = 255;
+
+	int image_size = (int)(image.height * image.width);
+	image.data.reserve(image_size);
+
+	//fill image data
+	for (int e = GRID_LENGTH_E - 1; e >= 0 ; e--) {
+	   for (int z = 0; z < GRID_LENGTH_Z; z++) {
```
nicovanduijn: Any particular reason you forward-iterate for azimuth but reverse-iterate for elevation? I'm sure there's a reason, but it's not obvious so a comment about it might be nice

jkflying: Otherwise the image comes out upside down, since it is written in byte order, not addressed with x/y

local_planner/src/nodes/local_planner.h
```diff
@@ -183,11 +184,13 @@ class LocalPlanner {
   void updateObstacleDistanceMsg(Histogram hist);
   void updateObstacleDistanceMsg();
   void create2DObstacleRepresentation(const bool send_to_fcu);
+  sensor_msgs::Image generateHistogramImage(Histogram& histogram);
 
  public:
   double h_FOV_ = 59.0;
   double v_FOV_ = 46.0;
   Box histogram_box_;
+  sensor_msgs::Image histogram_image_;
```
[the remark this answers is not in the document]
baumanta: it works the other way round. The main node queries for the image instead of the planner sends it out. Therefore it has to be stored until it is requested

PR #165, opened 2018-12-14
Add log replay functionality

baumanta:
This PR add a new launch file (sorry for even more launch files...) which enables you to play back a previously recorded rosbag and run the planner on the same pointcloud, position and goal data. This way you can compare your current version of the planner and its actions to the one recorded in the log.

This required two small changes to the planner:
- disable raise to goal altitude: I add a parameter which allow to disable the rising to goal altitude before starting with the planning. This is necessary because the log might not contain the arming position therefore the planner would never generate setpoint. This is also generally very useful to have for bench tests.
- listen to goal topic: I added a parameter which allows to listen to a topic which contain the goal location. This is necessary to grab the goal position from the bag file.

### DIFF
--- a/local_planner/src/nodes/local_planner_node.cpp
+++ b/local_planner/src/nodes/local_planner_node.cpp
@@ -683,7 +683,6 @@ void LocalPlannerNode::clickedGoalCallback(
 
 void LocalPlannerNode::updateGoalCallback(const visualization_msgs::MarkerArray& msg){
 	if(accept_goal_input_topic_ && msg.markers.size()>0){
-		std::cout<<"Getting a goal input\n";
 		goal_msg_.pose = msg.markers[0].pose;
 		new_goal_ = true;
 	}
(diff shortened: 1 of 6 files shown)


baumanta:
@jkflying the diasabling of the rising at start could also be interesting for your tests. That way you would not have to run the planner twice anymore

local_planner/src/nodes/local_planner_node.cpp
```diff
@@ -677,6 +681,14 @@ void LocalPlannerNode::clickedGoalCallback(
   goal_msg_.pose.position.z = local_planner_.getGoal().z;
 }
 
+void LocalPlannerNode::updateGoalCallback(const visualization_msgs::MarkerArray& msg){
+	if(accept_goal_input_topic_ && msg.markers.size()>0){
+		std::cout<<"Getting a goal input\n";
```
jkflying: Should this be a log instead?

PR #167, opened 2018-12-17
ReadMe: Mavros has to be installed from sources

baumanta:


TSC21:
@baumanta it has to in the next couple of weeks probably. After that, this isn't applicable anymore since the MAVROS will already have a release.

baumanta:
@TSC21 somebody had problems because they tried to install the package. I think in general it might probably be good to recommend installing from sources as the avoidance is still under development and we might have to add new messages and do not want to break it for everybody when we do. But it might not happen very often. What do you think is better?

What do you think @hamishwillee ?

mrivi:
I have asked @LorenzMeier if he can ask for a new mavros release since last one (v 0.27.0) was on November 12.

hamishwillee:
@TSC21 @mrivi Good points, good action. That said, I hate waiting on things that might not happen in a timely manner. Consider adding a note, and removing following the release.

> **Note:** MAVROS release 0.27.0 is incompatible with this software. A new release is expected imminently, but in the meantime install from sources.

Entirely up to you.

mrivi:
MAVROS release 0.28.0 is out, closing this

PR #169, opened 2018-12-19
Remove hard-coding of default goal_msg

nicovanduijn:
This replaces the hard-coded goal message and introduces some logic
to recognize if a proper goal has not been defined yet. (aka. it is NAN).

Drawback: we only check the pointcloud availability once the goal is set.

### DIFF
--- a/local_planner/src/nodes/local_planner_node.cpp
+++ b/local_planner/src/nodes/local_planner_node.cpp
@@ -111,9 +111,6 @@ LocalPlannerNode::~LocalPlannerNode() { delete server_; }
 void LocalPlannerNode::readParams() {
   // Parameter from launch file
   auto goal = local_planner_.getGoal();
-  nh_.param<double>("goal_x_param", goal.x, 9.0);
-  nh_.param<double>("goal_y_param", goal.y, 13.0);
-  nh_.param<double>("goal_z_param", goal.z, 3.5);
   nh_.param<bool>("disable_rise_to_goal_altitude", disable_rise_to_goal_altitude_, false);
   nh_.param<bool>("accept_goal_input_topic", accept_goal_input_topic_, false);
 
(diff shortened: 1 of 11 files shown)


local_planner/src/nodes/local_planner_node.h
```diff
@@ -94,7 +94,7 @@ class LocalPlannerNode {
   const ros::Duration pointcloud_timeout_hover_ = ros::Duration(0.4);
   const ros::Duration pointcloud_timeout_land_ = ros::Duration(10);
 
-  ros::Time last_wp_time_;
+  ros::Time last_pc_time_;
```
jkflying: `wp` stood for waypoint. What does `pc` stand for?

nicovanduijn: pointcloud. Should I expand?

local_planner/src/nodes/local_planner_node.cpp
```diff
@@ -111,9 +111,6 @@ LocalPlannerNode::~LocalPlannerNode() { delete server_; }
 void LocalPlannerNode::readParams() {
   // Parameter from launch file
   auto goal = local_planner_.getGoal();
-  nh_.param<double>("goal_x_param", goal.x, 9.0);
```
mrivi: can you remove the parameters `goal_*_param` from the launch files

local_planner/src/nodes/local_planner_node.h
jkflying: Unless it gets really long and difficult to work with, I prefer not to have abbreviations :+1:

local_planner/src/nodes/local_planner_node.cpp
nicovanduijn: Done. We need to keep the z-parameter though, since this would break the compatibility with RViz: the goal set when using the 2D Nav Goal button always has a z-coordinate of 0, which is why we don't want to overwrite it in its callback (`clickedGoalCallback`)

jkflying:
This would need to be re-thought now that there is the smoothing etc.

PR #176, opened 2019-01-15
Test for the common functions

AnnaDaiZH:
- added tests for the used common functions 
- deleted unused common functions
- fixed minor bugs in elevationAngletoIndex, azimuthAngletoIndex and velocityLinear
- changed some variable types from int to float, to enable more accurate calculation which also removes rounding error

### DIFF
--- a/local_planner/src/nodes/common.cpp
+++ b/local_planner/src/nodes/common.cpp
@@ -137,16 +137,6 @@ void normalize(geometry_msgs::Point& p) {
   }
 }
 
-double velocitySigmoid(double max_vel, double min_vel, double slope,
-                       double v_old, double elapsed) {
-  max_vel += 0.05;
-  min_vel -= 0.05;
-  v_old -= min_vel;
-  double t_old = -1.0 / slope * log((max_vel - min_vel) / v_old - 1.0);
-  double t_new = t_old + elapsed;
-  double speed = min_vel + (max_vel - min_vel) / (1.0 + exp(-slope * t_new));
-  return speed;
-}
 
 double velocityLinear(double max_vel, double min_vel, double slope,
                       double v_old, double elapsed) {
--- a/local_planner/src/nodes/common.cpp
+++ b/local_planner/src/nodes/common.cpp
@@ -15,8 +15,9 @@ float distance2DPolar(int e1, int z1, int e2, int z2) {
 
 float computeL2Dist(const geometry_msgs::Point& position,
                     const pcl::PointXYZ& xyz) {
-  return sqrt(pow(position.x - xyz.x, 2) + pow(position.y - xyz.y, 2) +
-              pow(position.z - xyz.z, 2));
+  return sqrt((position.x - xyz.x) * (position.x - xyz.x) +
+              (position.y - xyz.y) * (position.y - xyz.y) +
+              (position.z - xyz.z) * (position.z - xyz.z));
 }
 
 float distance3DCartesian(const geometry_msgs::Point& a,
@@ -29,7 +30,7 @@ float distance3DCartesian(const geometry_msgs::Point& a,
 geometry_msgs::Point fromPolarToCartesian(float e, float z, double radius,
                                           const geometry_msgs::Point& pos) {
   geometry_msgs::Point p;
-  p.x = pos.x + radius * cos(e * (M_PI / 180.f)) * sin(z * (M_PI / 180.f)); 
+  p.x = pos.x + radius * cos(e * (M_PI / 180.f)) * sin(z * (M_PI / 180.f));
   p.y = pos.y + radius * cos(e * (M_PI / 180.f)) * cos(z * (M_PI / 180.f));
   p.z = pos.z + radius * sin(e * (M_PI / 180.f));
 
@@ -41,8 +42,6 @@ double indexAngleDifference(float a, float b) {
                   std::abs(a - b + 360.f));
 }
 
-
-
 double elevationIndexToAngle(int e, double res) {
   return e * res + res / 2 - 90;
 }
@@ -52,17 +51,17 @@ double azimuthIndexToAngle(int z, double res) {
 }
 
 float azimuthAnglefromCartesian(const geometry_msgs::Point& position,
-                              const geometry_msgs::Point& origin) {
+                                const geometry_msgs::Point& origin) {
   return azimuthAnglefromCartesian(position.x, position.y, origin);
 }
 
 float azimuthAnglefromCartesian(double x, double y,
-                              const geometry_msgs::Point& pos) {
+                                const geometry_msgs::Point& pos) {
   return atan2(x - pos.x, y - pos.y) * 180.0 / M_PI;  //(-180. +180]
 }
 
 float elevationAnglefromCartesian(double x, double y, double z,
-                                const geometry_msgs::Point& pos) {
+                                  const geometry_msgs::Point& pos) {
   double den = sqrt((x - pos.x) * (x - pos.x) + (y - pos.y) * (y - pos.y));
   if (den == 0) {
     return 0;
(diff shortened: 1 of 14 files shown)


local_planner/src/nodes/common.cpp
```diff
@@ -179,31 +156,18 @@ void wrapAngleToPlusMinusPI(double& angle) {
...
 
   double vel;
-  if (std::abs(yaw_vel1) < std::abs(yaw_vel2)) {
-    vel = yaw_vel1;
-  } else {
-    vel = yaw_vel2;
-  }
-
+  // check which yaw direction is shorter
+  std::abs(yaw_vel1) <= std::abs(yaw_vel2)? vel = yaw_vel1: vel = yaw_vel2;
```
jkflying: This should be
` vel = (std::abs(yaw_vel1) <= std::abs(yaw_vel2))? yaw_vel1 : yaw_vel2;`

local_planner/src/nodes/common.cpp
```diff
@@ -75,46 +51,54 @@ double azimuthIndexToAngle(int z, double res) {
...
+int azimuthAngletoIndex(float z, int res) {  //[-180,180]
+  if(res <= 0.f || z < -180.f || z > 180.f){
+    return 0.f;
+  }
+  if (z == 180.f) {
     z = -180;
   }
   z += 180;
-  z = z + (res - (z % res));  //[-80,+90]
+  z = z + (res - ((int)z % res));  //[-80,+90]
```
nicovanduijn: c-style casts are not preferred in google style (https://google.github.io/styleguide/cppguide.html#Casting). use `static_cast<int>(z)`

local_planner/src/nodes/common.h
```diff
@@ -90,7 +90,19 @@ void normalize(geometry_msgs::Point& p);
 **/
 double velocityLinear(double max_vel, double slope,
                       double v_old, double elapsed);
+/**
+* @brief     wrappes the input angle in to plus minus PI space
+* @param[in] angle to be wrapped  [rad]
+* @returns   angle wrapped [rad]
```
nicovanduijn: this function doesn't return anything. instead write:
`@param[in,out] angle to be wrapped [rad]`

local_planner/test/test_common.cpp
```diff
@@ -26,3 +34,369 @@ TEST(Common, polar2DdistanceOnKnownPoints) {
...
+geometry_msgs::Point pos;
+pos.x = 0.f;
+pos.y = 0.f;
+pos.z = 0.f;
+
+std::vector<geometry_msgs::Point> pos_out;
+
+//WHEN: converting the point in polar CS to cartesian CS
+
+for(int i = 0; i<e.size(); i++){
```
nicovanduijn: use a range-based for loop

local_planner/test/test_common.cpp
```diff
@@ -26,3 +34,369 @@ TEST(Common, polar2DdistanceOnKnownPoints) {
...
+     pos_out.push_back(fromPolarToCartesian(e[i], z[3], radius[0], pos));
+}
+
+for(int i = 0; i<e.size(); i++){  
+       pos_out.push_back(fromPolarToCartesian(e[i], z[i], radius[1], pos));
```
nicovanduijn: This is dangerous! There's nothing guaranteeing you that `z` has the same dimensions as `e`, so `z[i]` could segfault

local_planner/test/test_common.cpp
```diff
@@ -26,3 +34,369 @@ TEST(Common, polar2DdistanceOnKnownPoints) {
...
+}
+
+
+//THEN: the cartesian coordinates are
+
+for(int i = 0; i<e.size(); i++){
```
nicovanduijn: Again, checking for the size of `e` when actually iterating over some other vector is not very safe. Consider something like `for (auto i : pos_out){`

local_planner/src/nodes/common.cpp
```diff
@@ -75,46 +52,54 @@ double azimuthIndexToAngle(int z, double res) {
...
                                 const geometry_msgs::Point& origin) {
   return elevationAnglefromCartesian(pos.x, pos.y, pos.z, origin);
 }
 
-int elevationAngletoIndex(int e, int res) {  //[-90,90]
-  if (e == 90) {
+int elevationAngletoIndex(float e, int res) {  //[-90,90]
+  // Do some input-checks
+  if(res <= 0.f || e < -90.f || e > 90.f){
+    return 0.f;
```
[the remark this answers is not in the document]
AnnaDaiZH: I changed the function to return a bool. Further i made the adaption of wrapping the elevation angle when input angle larger than 90/-90 but smaller than 180/-180. this needs to be tested first.

local_planner/src/nodes/common.cpp
```diff
@@ -75,46 +52,54 @@ double azimuthIndexToAngle(int z, double res) {
...
-  e = e + (res - (e % res));  //[-80,+90]
-  return e / res - 1;         //[0,17]
+  e = e + (res - ((int)e % res));  //[-80,+90]
+  return floor(e / res )- 1;         //[0,17]
 }
 
-int azimuthAngletoIndex(int z, int res) {  //[-180,180]
-  if (z == 180) {
+int azimuthAngletoIndex(float z, int res) {  //[-180,180]
+  if(res <= 0.f || z < -180.f || z > 180.f){
```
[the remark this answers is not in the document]
AnnaDaiZH: The problem is you cannot catch the case for invalid resolutions

local_planner/src/nodes/local_planner.cpp
```diff
@@ -263,9 +263,9 @@ void LocalPlanner::determineStrategy() {
         int n_occupied_cells = 0;
 
         int goal_e_angle =
-            elevationAnglefromCartesian(goal_, pose_.pose.position);
+            floor(elevationAnglefromCartesian(goal_, pose_.pose.position));
         int goal_z_angle =
-            azimuthAnglefromCartesian(goal_, pose_.pose.position);
+            floor(azimuthAnglefromCartesian(goal_, pose_.pose.position));
```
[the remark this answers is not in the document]
AnnaDaiZH: No, changed in new commit

local_planner/src/nodes/star_planner.cpp
```diff
@@ -111,8 +111,8 @@ double StarPlanner::treeCostFunction(int node_number) {
 }
 double StarPlanner::treeHeuristicFunction(int node_number) {
   geometry_msgs::Point node_position = tree_[node_number].getPosition();
-  int goal_z = azimuthAnglefromCartesian(goal_, node_position);
-  int goal_e = elevationAnglefromCartesian(goal_, node_position);
+  int goal_z = floor(azimuthAnglefromCartesian(goal_, node_position));
+  int goal_e = floor(elevationAnglefromCartesian(goal_, node_position));
```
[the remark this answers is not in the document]
AnnaDaiZH: The results are the same. Removed it in new commit

local_planner/src/nodes/common.cpp
baumanta: Agree, but I think we should leave that for now and open a new PR where we change it consistently in the whole code

local_planner/test/test_common.cpp
AnnaDaiZH: i have changed it to the range based for loops.  I just haven't pushed it yet.

AnnaDaiZH: the change will come in tomorrow

local_planner/src/nodes/common.cpp
```diff
@@ -152,23 +137,15 @@ void normalize(geometry_msgs::Point& p) {
...
 
-double velocityLinear(double max_vel, double min_vel, double slope,
+double velocityLinear(double max_vel,  double slope,
                       double v_old, double elapsed) {
-  v_old -= min_vel;
-  double t_old = v_old / slope;
-  double t_new = t_old + elapsed;
-  double speed = min_vel + t_new * slope;
+  double t_old = v_old / slope; // slope hard-coded as 1.f
+  double t_new = t_old + elapsed; //v_old + elapsed
```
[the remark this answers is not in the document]
AnnaDaiZH: removed

PR #178, opened 2019-01-15
Add unit tests for planner functions

mrivi:
Review only commit 12d8f1a. Tests are still a work in progress.

While writing the test for `FindFreeDirections` I found out that the indexes for blocked cells are not correctly wrapping around the histogram.
The issues are present when the moving window azimuth index is greater than the grid and when the moving window elevation is smaller than 0.

Before:
error_polar
After:
coorect_polar

### DIFF
--- a/local_planner/CMakeLists.txt
+++ b/local_planner/CMakeLists.txt
@@ -136,7 +136,7 @@ add_library(local_planner
   src/nodes/common.cpp
   src/nodes/rviz_world_loader.cpp
 )
-	
+
 ## Add cmake target dependencies of the library
 ## as an example, code may need to be generated before libraries
 ## either from message generation or dynamic reconfigure
@@ -214,7 +214,7 @@ if(CATKIN_ENABLE_TESTING)
 	                                             ${catkin_LIBRARIES}
 	                                             ${YAML_CPP_LIBRARIES})
 	endif()
-	
+
 	## Add folders to be run by python nosetests
 	# catkin_add_nosetests(test)
 endif()
--- a/local_planner/src/nodes/planner_functions.cpp
+++ b/local_planner/src/nodes/planner_functions.cpp
@@ -238,10 +238,10 @@ void combinedHistogram(bool& hist_empty, Histogram& new_hist,
 }
 
 // costfunction for every free histogram cell
-double costFunction(int e, int z, nav_msgs::GridCells &path_waypoints,
-                    const Eigen::Vector3f &goal,
-                    const Eigen::Vector3f &position,
-                    const Eigen::Vector3f &position_old, double goal_cost_param,
+double costFunction(int e, int z, nav_msgs::GridCells& path_waypoints,
+                    const Eigen::Vector3f& goal,
+                    const Eigen::Vector3f& position,
+                    const Eigen::Vector3f& position_old, double goal_cost_param,
                     double smooth_cost_param,
                     double height_change_cost_param_adapted,
                     double height_change_cost_param, bool only_yawed) {
--- a/local_planner/src/nodes/planner_functions.cpp
+++ b/local_planner/src/nodes/planner_functions.cpp
@@ -246,7 +246,11 @@ double costFunction(int e, int z, const nav_msgs::GridCells& path_waypoints,
                     double height_change_cost_param_adapted,
                     double height_change_cost_param, bool only_yawed) {
   double cost;
-  int waypoint_index = path_waypoints.cells.size();
+  int waypoint_index = path_waypoints.cells.size() - 1;
+  if (waypoint_index < 0) {
+    waypoint_index = 0;
+    path_waypoints.cells.push_back(position_old);
+  }
 
   double dist = (position - goal).norm();
   double dist_old = (position_old - goal).norm();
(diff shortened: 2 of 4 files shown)


nicovanduijn:
Love the visualization. What does the red color signify? Any reason there is no red 1 in the bottom right? (center of that blue cluster)

mrivi:
red is the rejected cell, the one to which the obstacle was mapped to. 
blue is a blocked cell. It's with in the safety radius of a rejected cell. 
green is a candidate cell which is free from obstacles

baumanta:
@mrivi yes, you are right. There is also a bug in the common function cartesian to polar and polar to cartesian. Those functions are not doing what they are supposed to. I plan on fixing that ASAP.

mrivi:
@baumanta ok, I'll finish the tests to make this PR complete. We should fly tests these fixes, right?

jkflying:
Looks good. Just a general comment, shouldn't the safety radius get wider as it gets higher/lower in the histogram, since it is in polar coordinates? Something like having a width of `radius/cos(elevation_angle)`?

nicovanduijn:
I'm still not sure I understand how this histogram works. Particularly:
- It makes sense to wrap horizontally, since -180 is  the same azimuth direction as +180, but I don't understand why there is a wrapping in elevation. An obstacle above shouldn't wrap to an obstacle below the drone, right?
- The red 1 seems to be duplicated on the "left" obstacle but missing on the right one. Almost makes it look like the wrapping across the top happens diagonally
- I agree with julian that the radius should depend on elevation. Think about it like this: if the obstacle is straight above, it should cover the entire top row of the histogram (all azimuth angles)

mrivi:
There is still a bug in the wrapping elevation. I have fixed it such that it wraps 180 in azimuth. Here's an example with a resolution of 20deg
screenshot from 2019-01-16 15-09-24

But there are still some corner cases. I am trying to fix those.

nicovanduijn:
Oh, now I get it. You're totally right. I thought the wrapping was top-to-bottom, which of course wouldn't make sense. You just happened to show two obstacles in the same azimuth direction, which mislead me a bit.

mrivi:
It should be working correctly now

local_planner/CMakeLists.txt
```diff
@@ -199,13 +199,14 @@ if(CATKIN_ENABLE_TESTING)
 	catkin_add_gtest(${PROJECT_NAME}-test test/main.cpp
 	                                      test/test_example.cpp
 	                                      test/test_common.cpp
-	                                      test/test_local_planner.cpp)
+	                                      test/test_local_planner.cpp
+	                                      test/test_planner_functions.cpp)
```
jkflying: This needs a rebase, there are new tests in `test_planner_functions.cpp` in master now.

local_planner/src/nodes/planner_functions.cpp
```diff
@@ -399,19 +403,26 @@ void findFreeDirections(
 
           // Elevation index < 0
           if (i < 0 && j >= 0 && j < z_dim) {
-            a = -i;
-            b = z_dim - j - 1;
+            a = 0;
```
jkflying: This logic gets a bit complicated, hopefully when we take into account the widths we can make a simpler version, maybe by switching to cartesian coordinates and back.

local_planner/test/test_planner_functions.cpp
```diff
@@ -0,0 +1,560 @@
...
+  calculateFOV(h_fov, v_fov, z_FOV_idx_z_greater_grid_length, e_FOV_min,
+               e_FOV_max, yaw_z_greater_grid_length, pitch);
+  calculateFOV(h_fov, v_fov, z_FOV_idx_z_max_greater_grid, e_FOV_min, e_FOV_max,
+               yaw_z_max_greater_grid, pitch);
+  calculateFOV(h_fov, v_fov, z_FOV_idx3_z_min_smaller_zero, e_FOV_min,
+               e_FOV_max, yaw_z_min_smaller_zero, pitch);
+  calculateFOV(h_fov, v_fov, z_FOV_idx_z_smaller_zero, e_FOV_min, e_FOV_max,
+               yaw_z_smaller_zero, pitch);
+
+  // THEN:
```
jkflying: Could you give a little explanation for what the checks do here? What are the different cases?

local_planner/test/test_planner_functions.cpp
```diff
@@ -78,3 +79,556 @@ TEST(PlannerFunctions, generateNewHistogramSpecificCells) {
...
+  findFreeDirections(histogram, safety_radius, path_candidates, path_selected,
+                     path_rejected, path_blocked, path_waypoints,
+                     cost_path_candidates, goal, position, position_old,
+                     goal_cost_param, smooth_cost_param,
+                     height_change_cost_param_adapted, height_change_cost_param,
+                     only_yawed, resolution_alpha);
+
+  // THEN: we should have one rejected cell and the 8 neightbooring cells
+  // blocked
+  for (int i = 0, j = 0; i < path_blocked.cells.size(), j < blocked.size();
```
jkflying: These test blocks seem to be used a lot, maybe they could be put in a function at the top of the file?

mrivi:
I think this PR needs outdoor tests before being merged

mrivi:
How should we go about merging this one? My testing capabilities are quite limited since I cannot run the simulation without timeouts on my laptop.

jkflying:
Why don't you rebase this, and I'll test it in SITL for you tomorrow.

mrivi:
@jkflying did you get a chance to test it?

jkflying:
SITL performance looks the same as master, feel free to merge.

mrivi:
@jkflying thank you!

PR #182, opened 2019-01-17
Cartesian polar conversion

baumanta:
Fix bug in calculation of elevation angle and add tests to further verify the conversions are correct


local_planner/src/nodes/common.cpp
```diff
@@ -63,11 +63,7 @@ float azimuthAnglefromCartesian(double x, double y,
 float elevationAnglefromCartesian(double x, double y, double z,
                                   const geometry_msgs::Point& pos) {
   double den = sqrt((x - pos.x) * (x - pos.x) + (y - pos.y) * (y - pos.y));
-  if (den == 0) {
-    return 0;
-  } else {
-    return atan((z - pos.z) / den) * 180.0 / M_PI;  //(-90.+90)
-  }
+  return atan2(z - pos.z, den) * 180.0 / M_PI;  //(-90.+90)
```
nicovanduijn: Make sure to also adjust the comments in the header files to reflect this change in functionality!

PR #183, opened 2019-01-21
run local_planner tests on travis

mrivi:


nicovanduijn:
Yeaiiii, we have C.I.

PR #185, opened 2019-01-21
Reduce includes

jkflying:
Reduce the number of un-needed includes and chained dependencies, which should help to keep compile times reasonable as the codebase grows.

I also pulled the output from the local planner into a separate file so that not everything depends on the local planner when it doesn't need to.

### DIFF
--- a/local_planner/src/nodes/local_planner_node.cpp
+++ b/local_planner/src/nodes/local_planner_node.cpp
@@ -256,9 +256,9 @@ void LocalPlannerNode::updatePlannerInfo() {
   // update ground distance
   if (ros::Time::now() - ground_distance_msg_.header.stamp <
       ros::Duration(0.5)) {
-    local_planner_.ground_distance_ = ground_distance_msg_.bottom_clearance;
+    local_planner_->ground_distance_ = ground_distance_msg_.bottom_clearance;
   } else {
-    local_planner_.ground_distance_ = 2.0;  // in case where no range data is
+    local_planner_->ground_distance_ = 2.0;  // in case where no range data is
                                             // available assume vehicle is close
                                             // to ground
   }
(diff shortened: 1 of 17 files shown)


local_planner/src/nodes/local_planner_node.cpp
```diff
@@ -745,7 +761,7 @@ void LocalPlannerNode::publishGround() {
 }
 
 void LocalPlannerNode::printPointInfo(double x, double y, double z) {
-  geometry_msgs::PoseStamped drone_pos = local_planner_.getPosition();
+  geometry_msgs::PoseStamped drone_pos = local_planner_->getPosition();
   int beta_z = floor(
```
nicovanduijn: isn't there a function in common.cpp for this?

local_planner/src/nodes/planner_functions.cpp
```diff
@@ -29,20 +33,20 @@ double adaptSafetyMarginHistogram(double dist_to_closest_point,
...
     double min_cloud_size, double min_dist_backoff, Box histogram_box,
-    const Eigen::Vector3f &position, double min_realsense_dist) {
+    const Eigen::Vector3f& position, double min_realsense_dist) {
   cropped_cloud.points.clear();
   cropped_cloud.width = 0;
   distance_to_closest_point = HUGE_VAL;
   float distance;
   counter_backoff = 0;
 
   for (size_t i = 0; i < complete_cloud.size(); ++i) {
```
nicovanduijn: outer loop could be range-based as well

local_planner/src/nodes/planner_functions.cpp
```diff
@@ -117,11 +121,12 @@ void calculateFOV(double h_fov, double v_fov, std::vector<int> &z_FOV_idx,
...
-                        pcl::PointCloud<pcl::PointXYZ> reprojected_points,
-                        std::vector<double> reprojected_points_age,
-                        std::vector<double> reprojected_points_dist,
-                        geometry_msgs::PoseStamped position) {
+void propagateHistogram(
+    Histogram& polar_histogram_est,
+    const pcl::PointCloud<pcl::PointXYZ>& reprojected_points,
+    const std::vector<double>& reprojected_points_age,
+    const std::vector<double>& reprojected_points_dist,
+    geometry_msgs::PoseStamped position) {
```
nicovanduijn: could this be passed by reference? It's quite big actually, 7 doubles plus the header

local_planner/src/nodes/local_planner_node.cpp
```diff
@@ -488,13 +502,15 @@ void LocalPlannerNode::publishBox() {
   box.pose.position.x = drone_pos.pose.position.x;
   box.pose.position.y = drone_pos.pose.position.y;
   box.pose.position.z = drone_pos.pose.position.z;
+                        0.5 * (local_planner_->histogram_box_.zsize_up_ -
```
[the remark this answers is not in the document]
jkflying: Rebase issues... I thought I caught this.

local_planner/src/nodes/planner_functions.cpp
jkflying: done

local_planner/src/nodes/planner_functions.cpp
jkflying: done

local_planner/src/nodes/local_planner_node.cpp
jkflying: I'd rather not change that in this PR

PR #190, opened 2019-01-23
Pr add model parameter

AnnaDaiZH:
add subscription to mavros parameter values for model parameters
add model parameters struct and parse the mavros parameter values to the struct

### DIFF
--- a/local_planner/src/nodes/local_planner_node.cpp
+++ b/local_planner/src/nodes/local_planner_node.cpp
@@ -58,6 +58,7 @@ LocalPlannerNode::LocalPlannerNode() {
                                   &LocalPlannerNode::updateGoalCallback, this);
   distance_sensor_sub_ = nh_.subscribe(
       "/mavros/altitude", 1, &LocalPlannerNode::distanceSensorCallback, this);
+  px4_param_sub_ = nh_.subscribe("/mavros/param/param_value", 1, &LocalPlannerNode::px4ParamsCallback, this);
 
   world_pub_ = nh_.advertise<visualization_msgs::MarkerArray>("/world", 1);
   drone_pub_ = nh_.advertise<visualization_msgs::Marker>("/drone", 1);
@@ -651,6 +652,13 @@ void LocalPlannerNode::distanceSensorCallback(
     publishGround();
   }
 }
+
+void LocalPlannerNode::px4ParamsCallback(const mavros_msgs::Param& msg) {
+
+  std::cout << msg.param_id << " value " << msg.value.integer << " o " << msg.value.real << std::endl;
+
+}
+
 void LocalPlannerNode::publishGround() {
   geometry_msgs::PoseStamped drone_pos = local_planner_->getPosition();
   visualization_msgs::Marker plane;
--- a/local_planner/src/nodes/local_planner_node.cpp
+++ b/local_planner/src/nodes/local_planner_node.cpp
@@ -58,7 +58,8 @@ LocalPlannerNode::LocalPlannerNode() {
                                   &LocalPlannerNode::updateGoalCallback, this);
   distance_sensor_sub_ = nh_.subscribe(
       "/mavros/altitude", 1, &LocalPlannerNode::distanceSensorCallback, this);
-  px4_param_sub_ = nh_.subscribe("/mavros/param/param_value", 1, &LocalPlannerNode::px4ParamsCallback, this);
+  px4_param_sub_ = nh_.subscribe("/mavros/param/param_value", 1,
+                                 &LocalPlannerNode::px4ParamsCallback, this);
 
   world_pub_ = nh_.advertise<visualization_msgs::MarkerArray>("/world", 1);
   drone_pub_ = nh_.advertise<visualization_msgs::Marker>("/drone", 1);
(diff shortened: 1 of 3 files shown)


local_planner/src/nodes/local_planner_node.h
```diff
@@ -53,6 +54,38 @@ struct cameraData {
   bool received_;
 };
 
+
+
+struct ModelParameters {
+  float ekf2_rng_a_hmax = NAN;
```
nicovanduijn: Comment every parameter, while we're at it

local_planner/src/nodes/local_planner_node.h
```diff
@@ -53,6 +54,38 @@ struct cameraData {
...
+  float mpc_thr_max = NAN;
+  float mpc_thr_min = NAN;
+  float mpc_tiltmax_air = NAN;
+  float mpc_tko_speed = NAN; // Takeoff climb rate
+  float mpc_xy_cruise = NAN; // Maximum horizontal velocity in mission, capped by mpc_xy_vel_max
+  float mpx_xy_vel_max = NAN; // Maximum horizontal velocity in AUTO mode. 
+  float mpc_z_vel_max_dn = NAN;
+  float mpc_z_vel_max_up = NAN;
+
+
```
nicovanduijn: Lots of white space issues in many places. Run `./tools/fix_style.sh local_planner` please. (This should also be caught by C.I.)

local_planner/src/nodes/local_planner_node.h
```diff
@@ -53,6 +54,38 @@ struct cameraData {
...
+  float mpc_hold_max_z = NAN; // Maximum vertical velocity for which position hold is enabled
+  float mpc_jerk_max = NAN;
+  float mpc_jerk_min = NAN;
+  int mpc_pos_mode = NAN; //smooth position velocity  mpc_pos_mode = 3
+  float mpc_thr_max = NAN;
+  float mpc_thr_min = NAN;
+  float mpc_tiltmax_air = NAN;
+  float mpc_tko_speed = NAN; // Takeoff climb rate
+  float mpc_xy_cruise = NAN; // Maximum horizontal velocity in mission, capped by mpc_xy_vel_max
+  float mpx_xy_vel_max = NAN; // Maximum horizontal velocity in AUTO mode.
```
mrivi: `mpx_xy_vel_max` -> `mpc_xy_vel_max`
The comment doesn't seem to be correct. The parameter applies also to manual modes.

jkflying:
@mrivi Next step is to take the outputs of this to a super simple struct with accel, velocity, position limits, but depending on which controller is being used it changes what parameters are being used for the limits so we need to read all of them to begin with.

local_planner/src/nodes/local_planner_node.h
```diff
@@ -53,6 +54,38 @@ struct cameraData {
...
+  float mc_yawauto_max = NAN;
+  float mpc_acc_down_max = NAN; // Maximum vertical acceleration in velocity controlled modes down
+  float mpc_acc_hor = NAN; // Maximum horizontal acceleration for auto mode and maximum deceleration for manual mode
+  float mpc_acc_up_max = NAN;  // Maximum vertical acceleration in velocity controlled modes upward
+  float mpc_col_prev_d = NAN; // Minimum distance the vehicle should keep to all obstacles
+  float mpc_hold_max_xy = NAN; // Maximum horizontal velocity for which position hold is enabled (use 0 to disable check)
+  float mpc_hold_max_z = NAN; // Maximum vertical velocity for which position hold is enabled
+  float mpc_jerk_max = NAN;
+  float mpc_jerk_min = NAN;
+  int mpc_pos_mode = NAN; //smooth position velocity  mpc_pos_mode = 3
```
mrivi: should be `MPC_AUTO_MODE` ?

AnnaDaiZH: both could be useful. added mpc_auto_mode

local_planner/src/nodes/local_planner_node.h
AnnaDaiZH: copied the comment from https://dev.px4.io/en/advanced/parameter_reference.html

local_planner/src/nodes/local_planner_node.cpp
```diff
@@ -730,6 +732,64 @@ void LocalPlannerNode::distanceSensorCallback(
     publishGround();
   }
 }
+
+void LocalPlannerNode::px4ParamsCallback(const mavros_msgs::Param& msg) {
+  // collect all px4 parameters needed for model based trajectory planning
```
nicovanduijn: These comments should go in the header file. Try to use similar style to [here](https://github.com/PX4/avoidance/blob/9b3d5ed6b2b9bace15f60a0b24eec5a8b3a7da7a/local_planner/src/nodes/common.h#L15)

local_planner/src/nodes/local_planner_node.h
```diff
@@ -53,6 +54,51 @@ struct cameraData {
   bool received_;
 };
 
+struct ModelParameters {
+  float ekf2_rng_a_hmax = NAN;  // Maximum absolute altitude (height above
```
nicovanduijn: Clang ripped this apart... consider wrapping this entire block in 
```
// clang-format off
   ...
// clang-format on
```

jkflying:
Just that last suggestion from Nico for the clang-format, and I'm happy

local_planner/src/nodes/local_planner_node.h
```diff
@@ -53,6 +55,33 @@ struct cameraData {
...
+  float up_vel = 3.0f;   // Maximum vertical ascent velocity
+  float down_acc = 10.0f; // Maximum vertical acceleration in velocity controlled modes down
+  float down_vel = 1.0f; // Maximum vertical descent velocity
+  float xy_acc = 5.0f;  // Maximum horizontal acceleration for auto mode and
+                      // maximum deceleration for manual mode
+  float xy_vel = 1.0f;   // Desired horizontal velocity in mission
+  float takeoff_speed = 1.0f; // Takeoff climb rate
+  float land_speed = 0.7f;   // Landing descend rate
+  // limitations given by sensors
+  float distance_sensor_max_height = 5.0f;
```
jkflying: This is mixing concerns, we shouldn't care about distance sensors inside of the simulator, just the limitations. We can leave this in, but it probably won't be used.

PR #191, opened 2019-01-23
Consistent waypoint smoothing

jkflying:
This PR vastly simplifies the waypoint smoothing by replacing it with a critically damped PD controller which has a single tuning variable instead of the 2 variables before.

Flight tested a lot in SITL, it gives smoother flight lines but still responds quickly to obstacles. Also, doing the smoothing in 3D gets rid of the vertical jumps we were seeing before.

I set up a mission in SITL, here is the flight log analysis of current master:
https://logs.px4.io/plot_app?log=c299ab1c-227a-4d45-8494-aa4bb4aba861

And of the new smoothing:
https://logs.px4.io/plot_app?log=181ca20c-3f48-4d4f-942f-1597432e3a49

Notice particularly the roll and pitch angular rates are much cleaner, with only a few spikes caused by major obstacles encountered (ie. real signal). Also notice how the altitude setpoint is much smoother.


local_planner/src/nodes/waypoint_generator.cpp
```diff
@@ -208,40 +208,31 @@ void WaypointGenerator::reachGoalAltitudeFirst() {
...
+
+    const Eigen::Vector3f p =
+        (desired_location - smoothed_goto_location_) * P_constant;
+    const Eigen::Vector3f d =
+        (desired_velocity - smoothed_goto_velocity_) * D_constant;
+
+    smoothed_goto_velocity_ += (p + d) * dt;
+    smoothed_goto_location_ += smoothed_goto_velocity_ * dt;
+
+    output_.smoothed_goto_position = toPoint(smoothed_goto_location_);
```
nicovanduijn: It might involve some juggling around with geometry_msgs and Eigen vectors, but could we extend this logic to also smooth the setpoint in orientation (yaw) and/or yaw rates? Most of the jerky behavior that makes for a bad UX comes from twitches in yaw I think

local_planner/src/nodes/waypoint_generator.cpp
```diff
@@ -208,40 +208,31 @@ void WaypointGenerator::reachGoalAltitudeFirst() {
...
-  // velocity-dependent max jerk
-  if (min_jerk_limit_param_ > 0.001f) {
-    max_jerk *= vel_cur_xy.norm();
-    if (max_jerk < min_jerk_limit_param_) max_jerk = min_jerk_limit_param_;
-  }
-
-  if (jerk_diff.squaredNorm() > max_jerk * max_jerk && max_jerk > 0.001f) {
-    jerk_diff = max_jerk * jerk_diff.normalized();
-    vel_sp_xy = (jerk_diff * dt + accel_cur) * dt + vel_cur_xy;
+  if (smoothing_speed_ > 0.01) {
```
nicovanduijn: Maybe it's worth noting that this smoothing only works for position setpoints. At first sight, I thought it was also smoothing when using the velocity setpoints... I got confused by the name `smoothed_goto_velocity`. Can we add that function and clarifying comments, should be quite similar, no?

Also, I just realized that I don't understand the `struct waypointResult` altogether. there is a `position_waypoint`, `goto_position`, `adapted_goto_position` and `smoothed_goto_position`. Desperately needs some comments in the code, those all sound the same to me.

local_planner/src/nodes/waypoint_generator.cpp
```diff
@@ -208,40 +208,31 @@ void WaypointGenerator::reachGoalAltitudeFirst() {
...
+    const float P_constant = smoothing_speed_;
+    const float D_constant = 2.f * std::sqrt(P_constant);  // critically damped
+    const Eigen::Vector3f desired_location =
+        toEigen(output_.adapted_goto_position);
+    const Eigen::Vector3f desired_velocity = toEigen(curr_vel_.twist.linear);
```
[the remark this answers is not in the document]
jkflying: This is the desired velocity of the setpoint, not of the drone. The idea is that if the drone is moving, the setpoint should move too, since it should stay a constant distance in front of the drone. So we don't penalize movement of the setpoint in general, just movement relative to the drone.

local_planner/src/nodes/waypoint_generator.cpp
jkflying: Agreed, the yaw angular rates are still similarly noisy to the old one, while the pitch and roll improved a lot. Unfortunately I don't think there is much we can do at the smoothing side without impacting the performance of avoiding obstacles - since most of the actual signal in the unsmoothed setpoint movement is in yaw, that is where most of the noise still is. I think the biggest help here will be getting smoother outputs from the planner - a smoother can only do so much.

local_planner/src/nodes/waypoint_generator.cpp
jkflying: Here's a log of the same mission using velocity setpoints, it still seems to be smoothed, but travels a lot faster.
https://logs.px4.io/plot_app?log=d7ae64d4-c6d6-42f1-a60c-1f5a776c595f

jkflying: I will rename the variable to make it more clear

local_planner/src/nodes/waypoint_generator.cpp
jkflying: Yaw setpoint wasn't using the smoothed setpoint! Big improvement in yaw angular rates:
https://logs.px4.io/plot_app?log=fbc84d4f-c683-4582-bea5-1cc4c623bdaa

jkflying:
After the last 2 commits the yaw is much smoother, and it turns in the correct direction (and doesn't get stuck) when reaching waypoints.

See before and after the last 2 commits:
image

SITL log for latest commit:
https://logs.px4.io/plot_app?log=03f7d38e-12dd-44a8-834a-9df3981e8a0c

PR #193, opened 2019-01-24
Create struct for polar points

AnnaDaiZH:
before all polar points were defined as int e for elevation angle, int z for azimuth angle and double r for radius separately

with this PR

- PolarPoints has the member float e, float z and float r
- all conversion functions in common.cpp are changed to use the PolarPoint if needed
- implementation of the function in local_planner/
- no changes in the behaviour of the replaced functions

### DIFF
--- a/local_planner/src/nodes/common.cpp
+++ b/local_planner/src/nodes/common.cpp
@@ -33,6 +33,7 @@ double indexAngleDifference(float a, float b) {
 
 PolarPoint histogramIndexToPolar(int e, int z, int res, float radius) {
   PolarPoint p_pol;
+  // ALPHA_RES%2=0 as per definition, see histogram.h
   p_pol.e = static_cast<float>(e * res + res / 2 - 90);
   p_pol.z = static_cast<float>(z * res + res / 2 - 180);
   p_pol.r = radius;
--- a/local_planner/src/nodes/common.cpp
+++ b/local_planner/src/nodes/common.cpp
@@ -42,26 +42,7 @@ PolarPoint HistogramIndexToPolar(const int& e, const int& z, const int& res,
   return p_pol;
 }
 
-float azimuthAnglefromCartesian(const Eigen::Vector3f& position,
-                                const Eigen::Vector3f& origin) {
-  return azimuthAnglefromCartesian(position.x(), position.y(), origin);
-}
 
-float azimuthAnglefromCartesian(double x, double y,
-                                const Eigen::Vector3f& pos) {
-  return atan2(x - pos.x(), y - pos.y()) * (180.0 / M_PI);  //(-180. +180]
-}
-
-float elevationAnglefromCartesian(double x, double y, double z,
-                                  const Eigen::Vector3f& pos) {
-  double den = (Eigen::Vector2f(x, y) - pos.topRows<2>()).norm();
-  return atan2(z - pos.z(), den) * 180.0 / M_PI;  //(-90.+90)
-}
-
-float elevationAnglefromCartesian(const Eigen::Vector3f& pos,
-                                  const Eigen::Vector3f& origin) {
-  return elevationAnglefromCartesian(pos.x(), pos.y(), pos.z(), origin);
-}
 
 PolarPoint CartesianToPolar(const Eigen::Vector3f& pos,
                             const Eigen::Vector3f& origin) {
@@ -79,6 +60,10 @@ PolarPoint CartesianToPolar(double x, double y, double z,
   return p_pol;
 }
 
+
+Eigen::Vector2i PolarToHistogramIndex(const PolarPoint& p_pol, const int& res){
+  
+}
 int elevationAngletoIndex(float e, int res) {  //[-90,90]
   // TODO: wrap e to [-90, 90] to be sure input is valid such that this check is
   // not necessary anymore
--- a/local_planner/src/nodes/common.cpp
+++ b/local_planner/src/nodes/common.cpp
@@ -13,9 +13,7 @@ float distance2DPolar(const PolarPoint& p1, const PolarPoint& p2) {
   return sqrt((p1.e - p2.e) * (p1.e - p2.e) + (p1.z - p2.z) * (p1.z - p2.z));
 }
 
-// transform polar coordinates into Cartesian coordinates
-
-Eigen::Vector3f PolarToCartesian(const PolarPoint& p_pol,
+Eigen::Vector3f polarToCartesian(const PolarPoint& p_pol,
                                  const geometry_msgs::Point& pos) {
   Eigen::Vector3f p;
   p.x() =
@@ -33,31 +31,30 @@ double indexAngleDifference(float a, float b) {
                   std::abs(a - b + 360.f));
 }
 
-PolarPoint HistogramIndexToPolar(int e, int z, int res, const float& radius) {
-  PolarPoint p_pol = {};
-  p_pol.e = e * res + res / 2 - 90;
-  p_pol.z = z * res + res / 2 - 180;
+PolarPoint histogramIndexToPolar(int e, int z, int res, float radius) {
+  PolarPoint p_pol;
+  p_pol.e = static_cast<float>(e * res + res / 2 - 90);
+  p_pol.z = static_cast<float>(z * res + res / 2 - 180);
   p_pol.r = radius;
   return p_pol;
 }
 
-PolarPoint CartesianToPolar(const Eigen::Vector3f& pos,
+PolarPoint cartesianToPolar(const Eigen::Vector3f& pos,
                             const Eigen::Vector3f& origin) {
-  return CartesianToPolar(pos.x(), pos.y(), pos.z(), origin);
+  return cartesianToPolar(pos.x(), pos.y(), pos.z(), origin);
 }
-PolarPoint CartesianToPolar(double x, double y, double z,
+PolarPoint cartesianToPolar(double x, double y, double z,
                             const Eigen::Vector3f& pos) {
-  PolarPoint p_pol = {};
-  p_pol.z = atan2(x - pos.x(), y - pos.y()) * (180.0 / M_PI);  //(-180. +180]
-
+  PolarPoint p_pol;
   double den = (Eigen::Vector2f(x, y) - pos.topRows<2>()).norm();
-  p_pol.e = atan2(z - pos.z(), den) * 180.0 / M_PI;  //(-90.+90)
+  p_pol.e = atan2(z - pos.z(), den) * 180.0 / M_PI;            //(-90.+90)
+  p_pol.z = atan2(x - pos.x(), y - pos.y()) * (180.0 / M_PI);  //(-180. +180]
   p_pol.r = sqrt((x - pos.x()) * (x - pos.x()) + (y - pos.y()) * (y - pos.y()) +
                  (z - pos.z()) * (z - pos.z()));
   return p_pol;
 }
 
-Eigen::Vector2i PolarToHistogramIndex(const PolarPoint& p_pol, int res) {
+Eigen::Vector2i polarToHistogramIndex(const PolarPoint& p_pol, int res) {
   // TODO change logic here, as 0,0 are valid index
   Eigen::Vector2i ev2(0, 0);
   float e = p_pol.e;
--- a/local_planner/src/nodes/common.cpp
+++ b/local_planner/src/nodes/common.cpp
@@ -9,19 +9,21 @@
 #include <tf/transform_listener.h>
 namespace avoidance {
 
-
 float distance2DPolar(const PolarPoint& p1, const PolarPoint& p2) {
-  return sqrt((p1.e-p2.e)*(p1.e-p2.e) + (p1.z-p2.z)*(p1.z-p2.z));
+  return sqrt((p1.e - p2.e) * (p1.e - p2.e) + (p1.z - p2.z) * (p1.z - p2.z));
 }
-
 
 // transform polar coordinates into Cartesian coordinates
 
-Eigen::Vector3f fromPolarToCartesian(const PolarPoint& p_pol,
-                                     const geometry_msgs::Point& pos) {
+Eigen::Vector3f PolarToCartesian(const PolarPoint& p_pol,
+                                 const geometry_msgs::Point& pos) {
   Eigen::Vector3f p;
-  p.x() = pos.x + p_pol.r * cos(p_pol.e * (M_PI / 180.f)) * sin(p_pol.z * (M_PI / 180.f));
-  p.y() = pos.y + p_pol.r * cos(p_pol.e * (M_PI / 180.f)) * cos(p_pol.z * (M_PI / 180.f));
+  p.x() =
+      pos.x +
+      p_pol.r * cos(p_pol.e * (M_PI / 180.f)) * sin(p_pol.z * (M_PI / 180.f));
+  p.y() =
+      pos.y +
+      p_pol.r * cos(p_pol.e * (M_PI / 180.f)) * cos(p_pol.z * (M_PI / 180.f));
   p.z() = pos.z + p_pol.r * sin(p_pol.e * (M_PI / 180.f));
 
   return p;
@@ -31,12 +33,13 @@ double indexAngleDifference(float a, float b) {
                   std::abs(a - b + 360.f));
 }
 
-double elevationIndexToAngle(int e, double res) {
-  return e * res + res / 2 - 90;
-}
-
-double azimuthIndexToAngle(int z, double res) {
-  return z * res + res / 2 - 180;
+PolarPoint HistogramIndexToPolar(const int& e, const int& z, const int& res,
+                                 const float& radius) {
+  PolarPoint p_pol = {};
+  p_pol.e = e * res + res / 2 - 90;
+  p_pol.z = z * res + res / 2 - 180;
+  p_pol.r = radius;
+  return p_pol;
 }
 
 float azimuthAnglefromCartesian(const Eigen::Vector3f& position,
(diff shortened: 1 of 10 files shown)


local_planner/src/nodes/common.cpp
```diff
@@ -9,82 +9,76 @@
...
                   std::abs(a - b + 360.f));
 }
 
-double elevationIndexToAngle(int e, double res) {
-  return e * res + res / 2 - 90;
-}
-
-double azimuthIndexToAngle(int z, double res) {
-  return z * res + res / 2 - 180;
+PolarPoint HistogramIndexToPolar(int e, int z, int res, const float& radius) {
```
nicovanduijn: float doesn't need to be passed by reference

local_planner/src/nodes/common.cpp
```diff
@@ -9,82 +9,76 @@
...
+  PolarPoint p_pol = {};
+  p_pol.e = e * res + res / 2 - 90;
```
nicovanduijn: are you consciously doing integer arithmetic here? Consider casting first, because integer arithmetic can yield some non-obvious results... Otherwise, put a `static_cast<float>()` around the whole operation such that you're not doing so implicitly.

local_planner/src/nodes/common.cpp
```diff
@@ -9,82 +9,76 @@
...
 }
 
-double elevationIndexToAngle(int e, double res) {
-  return e * res + res / 2 - 90;
-}
-
-double azimuthIndexToAngle(int z, double res) {
-  return z * res + res / 2 - 180;
+PolarPoint HistogramIndexToPolar(int e, int z, int res, const float& radius) {
+  PolarPoint p_pol = {};
```
nicovanduijn: Consider giving this struct a constructor, might make lines like this prettier

local_planner/src/nodes/common.cpp
```diff
@@ -9,82 +9,76 @@
...
-  return sqrt(pow((e1 - e2), 2) + pow((z1 - z2), 2));
+float distance2DPolar(const PolarPoint& p1, const PolarPoint& p2) {
+  return sqrt((p1.e - p2.e) * (p1.e - p2.e) + (p1.z - p2.z) * (p1.z - p2.z));
 }
 
 // transform polar coordinates into Cartesian coordinates
-Eigen::Vector3f fromPolarToCartesian(float e, float z, double radius,
-                                     const geometry_msgs::Point& pos) {
+
+Eigen::Vector3f PolarToCartesian(const PolarPoint& p_pol,
```
nicovanduijn: lower-case first letter of member function name `polarToCartesian`

local_planner/src/nodes/common.cpp
```diff
@@ -9,82 +9,76 @@
...
+  p_pol.e = e * res + res / 2 - 90;
+  p_pol.z = z * res + res / 2 - 180;
+  p_pol.r = radius;
+  return p_pol;
 }
 
-float azimuthAnglefromCartesian(const Eigen::Vector3f& position,
-                                const Eigen::Vector3f& origin) {
-  return azimuthAnglefromCartesian(position.x(), position.y(), origin);
+PolarPoint CartesianToPolar(const Eigen::Vector3f& pos,
```
nicovanduijn: `cartesianToPolar`

local_planner/src/nodes/common.h
```diff
@@ -10,58 +10,53 @@
...
 namespace avoidance {
 
-float distance2DPolar(int e1, int z1, int e2, int z2);
+struct PolarPoint {
+  float e;
+  float z;
+  float r;
+};
+
+float distance2DPolar(const PolarPoint& p1, const PolarPoint& p2);
```
nicovanduijn: Since this function is quite obvious, it doesn't need lots of comments, but a `@brief` with one line would be nice

local_planner/src/nodes/common.h
```diff
@@ -10,58 +10,53 @@
...
 
 /**
 * @brief     Convertes the point from polar CS to cartesian CS
-* @param[in] e elevation angle (-90, 90]
-* @param[in] z azimuth angle (-180,180]
-* @param[in] radius
+* @param[in] p_pol polar point
```
nicovanduijn: These comments are not so correct anymore. When I see a function `polarToCartesian()`, I expect it to take in a polar point, convert it to cartesian and return that. Why are there two arguments? This should be answered by the comments and preferably obvious from the function name. What the function actually does is add a polar unit vector to a cartesian point, returning the resulting 3D point in cartesian format. (Right? I may even be wrong here!)

local_planner/src/nodes/common.h
```diff
@@ -10,58 +10,53 @@
...
+                                 const geometry_msgs::Point& pos);
 double indexAngleDifference(float a, float b);
 
-double elevationIndexToAngle(int e, double res);
-double azimuthIndexToAngle(int z, double res);
+PolarPoint HistogramIndexToPolar(int e, int z, int res, const float& radius);
 
 /**
-* @brief     Compute the bearing angle of a given vector relative to an
+* @brief     Compute the polar point to an
```
nicovanduijn: Careful... The brief is no longer really explaining much. Compute the polar point to an origin? What?

local_planner/src/nodes/local_planner.cpp
```diff
@@ -404,7 +401,10 @@ void LocalPlanner::updateObstacleDistanceMsg() {
 void LocalPlanner::reprojectPoints(Histogram histogram) {
   double n_points = 0;
   double dist, age;
+  float half_res = ALPHA_RES / 2;
```
nicovanduijn: are you aware of what this does? Careful...!

local_planner/src/nodes/local_planner.cpp
```diff
@@ -416,23 +416,29 @@ void LocalPlanner::reprojectPoints(Histogram histogram) {
...
         n_points++;
+        for (auto &i : p_pol) {
+          i.r = histogram.get_dist(e, z);
+          i = HistogramIndexToPolar(e, z, ALPHA_RES, i.r);
+        }
         // transform from array index to angle
-        double beta_e = elevationIndexToAngle(e, ALPHA_RES);
-        double beta_z = azimuthIndexToAngle(z, ALPHA_RES);
+
+        // float beta_e = elevationIndexToAngle(e, ALPHA_RES);
```
nicovanduijn: delete commented code

local_planner/src/nodes/planner_functions.cpp
```diff
@@ -248,14 +244,20 @@ double costFunction(int e, int z, const nav_msgs::GridCells& path_waypoints,
...
   int waypoint_index = path_waypoints.cells.size();
 
-  double dist = (position - goal).norm();
-  double dist_old = (position_old - goal).norm();
-  Eigen::Vector3f candidate_goal =
-      fromPolarToCartesian(e, z, dist, toPoint(position));
+  float dist = (position - goal).norm();
+  float dist_old = (position_old - goal).norm();
+  PolarPoint p_pol = {};
+  p_pol.e = static_cast<float>(e);
```
nicovanduijn: no need to cast anymore

local_planner/src/nodes/common.h
```diff
@@ -10,58 +10,53 @@
...
 /**
-* @brief     Checks if the computed histogram index given an elevation angle and
-*resolution is valid
-* @param[in] e Elevation angle in degrees
+* @brief     compute polar point to histogram index
+* @param[in] p_pol with elevation, azimuth angle and radius
 * @param[in] res resolution of the histogram in degrees
-* @param[in] index Index of the histogram cell for the given elevation
-* @returns   true or
+* @param[out] eigenvector with x()=azimuth and y()=elevation
```
baumanta: eigenvector is a bit loaded. I would just write vector, the format is specified in the function

local_planner/src/nodes/local_planner.cpp
baumanta: half_res needs to be an integer. Otherwise it looks fine to me. The resolution is hard coded and it needs to be dividable by two, otherwise we have a lot more problems. I will write a test for valid resolutions in another PR.

local_planner/src/nodes/planner_functions.cpp
baumanta: you could also make to cost function take a polar point (reference) as an input. That way you would have less arguments.

local_planner/src/nodes/planner_functions.cpp
```diff
@@ -496,10 +504,9 @@ bool getDirectionFromTree(
           (1.f - l_frac) * toEigen(path_node_positions[wp_idx - 1]) +
           l_frac * toEigen(path_node_positions[wp_idx]);
 
-      int wp_e = floor(elevationAnglefromCartesian(mean_point, position));
-      int wp_z = floor(azimuthAnglefromCartesian(mean_point, position));
+      PolarPoint wp_pol = CartesianToPolar(mean_point, position);
 
-      p = Eigen::Vector3f(wp_e, wp_z, 0.f);
+      p = Eigen::Vector3f(wp_pol.e, wp_pol.z, 0.f);
```
baumanta: I would change the function to output a polar point. Here you mix up the types again. It would be nicer if wp_pol would be the output, such that also seen from the outside, you know that you are getting the waypoint in polar coordinates, not in cartesian ones.

local_planner/src/nodes/local_planner.cpp
```diff
@@ -123,11 +123,13 @@ void LocalPlanner::runPlanner() {
   // visualization of FOV in RViz
   initGridCells(FOV_cells_);
   geometry_msgs::Point p;
+  PolarPoint p_pol = {};
```
jkflying: If you don't want to put data in it you don't need the ` = {}`. However, you aren't accessing it outside the loop, so just declare it below inside the loop when you set it.

local_planner/src/nodes/local_planner.cpp
```diff
@@ -404,7 +401,10 @@ void LocalPlanner::updateObstacleDistanceMsg() {
 void LocalPlanner::reprojectPoints(Histogram histogram) {
   double n_points = 0;
   double dist, age;
+  float half_res = ALPHA_RES / 2;
   Eigen::Vector3f temp_array[4];
+
+  std::vector<PolarPoint> p_pol(4);
```
jkflying: Rather make this a matrix than a vector, since you know beforehand how many you actually need. Or better yet you can use a std::array

local_planner/src/nodes/local_planner_node.cpp
```diff
@@ -359,9 +359,12 @@ void LocalPlannerNode::initMarker(visualization_msgs::MarkerArray* marker,
...
     m.id = i + 1;
     m.action = visualization_msgs::Marker::ADD;
+    PolarPoint p_pol = {};
+    p_pol.e = path.cells[i].x;
+    p_pol.z = path.cells[i].y;
+    p_pol.r = 1.0;
     geometry_msgs::Point p =
-        toPoint(fromPolarToCartesian((int)path.cells[i].x, (int)path.cells[i].y,
-                                     1.0, drone_pos.pose.position));
+        toPoint(PolarToCartesian(p_pol, drone_pos.pose.position));
```
jkflying: Nothing else uses `p`, so just assign it straight to `m.pose.position`

local_planner/src/nodes/planner_functions.cpp
```diff
@@ -289,13 +291,15 @@ double costFunction(int e, int z, const nav_msgs::GridCells& path_waypoints,
...
                                 const Histogram& input_hist) {
-  int vertical_FOV_range_sensor = 20;
-  int lower_index = elevationAngletoIndex(
-      -(float)(vertical_FOV_range_sensor / 2.0), ALPHA_RES);
-  int upper_index = elevationAngletoIndex(
-      (float)(vertical_FOV_range_sensor / 2.0), ALPHA_RES);
-
-  for (int e = lower_index; e <= upper_index; e++) {
+  float vertical_FOV_range_sensor = 20.0;
+  PolarPoint p_pol_lower = {};
```
jkflying: Put the below initialization straight inside, with a dummy variable for the z value, so we don't have uninitialized members of our struct.

local_planner/src/nodes/planner_functions.cpp
```diff
@@ -385,25 +390,28 @@ void findFreeDirections(
...
         path_rejected.cells.push_back(p);
       } else {
-        p.x = elevationIndexToAngle(e, resolution_alpha);
-        p.y = azimuthIndexToAngle(z, resolution_alpha);
-        p.z = 0.0;
+        p_pol = HistogramIndexToPolar(e, z, resolution_alpha, 0.0);
+        p.x = p_pol.e;
+        p.y = p_pol.z;
+        p.z = p_pol.r;
         path_blocked.cells.push_back(p);
```
jkflying: If we're If we are sticking polar points into the `path_blocked.cells`, maybe it should be a vector of `PolarPoint`s?

local_planner/src/nodes/planner_functions.cpp
jkflying: The `p` here should also be a  `PolarPoint`, right?

local_planner/src/nodes/waypoint_generator.cpp
```diff
@@ -34,9 +34,12 @@ void WaypointGenerator::calculateWaypoint() {
       break;
     }
     case costmap: {
-      output_.goto_position = toPoint(fromPolarToCartesian(
-          planner_info_.costmap_direction_e, planner_info_.costmap_direction_z,
-          1.0, planner_info_.pose.pose.position));
+      PolarPoint p_pol = {};
```
jkflying: Initialize straight inside, not afterwards.

local_planner/src/nodes/waypoint_generator.cpp
```diff
@@ -53,8 +56,12 @@ void WaypointGenerator::calculateWaypoint() {
       if (tree_available && (planner_info_.obstacle_ahead || dist_goal > 4.0) &&
           since_last_path < ros::Duration(5)) {
         ROS_DEBUG("[WG] Use calculated tree\n");
-        output_.goto_position = toPoint(fromPolarToCartesian(
-            p.x(), p.y(), 1.0, planner_info_.pose.pose.position));
+        PolarPoint p_pol = {};
```
jkflying: Initialize directly

local_planner/test/test_common.cpp
```diff
@@ -6,22 +6,29 @@ using namespace avoidance;
 
 TEST(Common, polar2DdistanceSameIsZero) {
   // GIVEN: two identical points
-  const int e = 5, z = 9;
+  PolarPoint p = {};
```
jkflying: Initialize directly

local_planner/test/test_common.cpp
```diff
@@ -6,22 +6,29 @@ using namespace avoidance;
...
 
   // THEN: the distance should be zero
   EXPECT_FLOAT_EQ(0.f, dist);
 }
 
 TEST(Common, polar2DdistanceOnKnownPoints) {
   // GIVEN: two points
-  const int e1 = 5, z1 = 9;
-  const int e2 = 50, z2 = 39;
+  PolarPoint p1 = {};
```
jkflying: Initialize directly

local_planner/test/test_common.cpp
```diff
@@ -127,32 +133,38 @@ TEST(Common, elevationAnglefromCartesian) {
 
 TEST(Common, elevationAngletoIndex) {
```
jkflying: I don't think you need to test the elevation angle and the azimuth angle to index separately. Check them together, in a single test, so you don't have half-initialized structs everywhere.

local_planner/test/test_common.cpp
```diff
@@ -232,13 +251,19 @@ TEST(Common, fromPolarToCartesian) {
   // WHEN: converting the point in polar CS to cartesian CS
 
   for (int i = 0; i < n; i++) {
-    pos_out.push_back(
-        fromPolarToCartesian(e[i], z[3], radius[0], toPoint(pos)));
+    PolarPoint p_pol = {};
```
jkflying: Initialize directly

local_planner/test/test_common.cpp
```diff
@@ -232,13 +251,19 @@ TEST(Common, fromPolarToCartesian) {
...
+    p_pol.e = e[i];
+    p_pol.z = z[3];
+    p_pol.r = radius[0];
+    pos_out.push_back(PolarToCartesian(p_pol, toPoint(pos)));
   }
 
   for (int i = 0; i < n; i++) {
-    pos_out.push_back(
-        fromPolarToCartesian(e[i], z[i], radius[1], toPoint(pos)));
+    PolarPoint p_pol = {};
```
jkflying: Initialize directly

local_planner/test/test_common.cpp
```diff
@@ -279,26 +304,29 @@ TEST(Common, PolarToCatesianToPolar) {
   // cartesian and back again
   for (float e = -90.f; e <= 90.f; e = e + 3.f) {
     for (float z = -180.f; z <= 180.f; z = z + 6.f) {
-      Eigen::Vector3f p_cartesian =
-          fromPolarToCartesian(e, z, radius, toPoint(pos));
-      float z_new = azimuthAnglefromCartesian(p_cartesian, pos);
-      float e_new = elevationAnglefromCartesian(p_cartesian, pos);
+      PolarPoint p_pol = {};
```
jkflying: Initialize directly

local_planner/test/test_planner_functions.cpp
```diff
@@ -33,17 +33,19 @@ TEST(PlannerFunctions, generateNewHistogramSpecificCells) {
...
   std::vector<Eigen::Vector3f> middle_of_cell;
 
-  for (int i = 0; i < e_angle_filled.size(); i++) {
-    for (int j = 0; j < z_angle_filled.size(); j++) {
-      middle_of_cell.push_back(fromPolarToCartesian(e_angle_filled[i],
-                                                    z_angle_filled[j], distance,
-                                                    location.pose.position));
+  for (auto i : e_angle_filled) {
+    for (auto j : z_angle_filled) {
+      PolarPoint p_pol = {};
```
jkflying: Initialize directly

local_planner/test/test_planner_functions.cpp
```diff
@@ -63,8 +65,11 @@ TEST(PlannerFunctions, generateNewHistogramSpecificCells) {
   std::vector<int> e_index;
   std::vector<int> z_index;
   for (int i = 0; i < e_angle_filled.size(); i++) {
-    e_index.push_back(elevationAngletoIndex((int)e_angle_filled[i], ALPHA_RES));
-    z_index.push_back(azimuthAngletoIndex((int)z_angle_filled[i], ALPHA_RES));
+    PolarPoint p_pol = {};
```
jkflying: Initialize directly

local_planner/src/nodes/planner_functions.cpp
```diff
@@ -252,14 +248,15 @@ double costFunction(int e, int z, nav_msgs::GridCells& path_waypoints,
...
 
-  double dist = (position - goal).norm();
-  double dist_old = (position_old - goal).norm();
-  Eigen::Vector3f candidate_goal =
-      fromPolarToCartesian(e, z, dist, toPoint(position));
+  float dist = (position - goal).norm();
+  float dist_old = (position_old - goal).norm();
+  p_pol.r = dist;
+  Eigen::Vector3f candidate_goal = polarToCartesian(p_pol, toPoint(position));
+  PolarPoint p_pol_old(path_waypoints.cells[waypoint_index - 1].x,
```
jkflying: This is reading from an invalid index if the waypoint_index is 0. I think we should get rid of the -1.

PR #195, opened 2019-01-25
Cost matrix: distance dependent cost

baumanta:
This PR changes the The histogram format and cost function completely. Here a few points that changed:

- The histogram is now stored as an Eigen::Matrix
- The histogram now contains only a distance and age layer, the binary is dropped
- The cost of each histogram cell is stored in a cost matrix
- A new cost-function is used, which contains the obstacle distance in the cost value
- To smooth the cost matrix (former safety radius) the matrix is padded (wrapping correctly around elevation and azimuth). and then a max median filter is used on it.
- A lot of tests for the new functions

We definitely have to flight test this before merging. It might fly a bit worse than before, as I did not yet spend too much time on re-tuning the costmatrix smoothing and costfunction.

### DIFF
--- a/local_planner/src/nodes/candidate_direction.h
+++ b/local_planner/src/nodes/candidate_direction.h
@@ -3,11 +3,11 @@
 namespace avoidance {
 
 struct candidateDirection {
-  double cost;
-  double elevation_angle;
-  double azimuth_angle;
+  float cost;
+  float elevation_angle;
+  float azimuth_angle;
 
-  candidateDirection(double c, double e, double z)
+  candidateDirection(float c, float e, float z)
       : cost(c), elevation_angle(e), azimuth_angle(z){};
 
   bool operator<(const candidateDirection& y) const { return cost < y.cost; }
--- a/local_planner/src/nodes/candidate_direction.h
+++ b/local_planner/src/nodes/candidate_direction.h
@@ -1,0 +1,17 @@
+#pragma once
+
+namespace avoidance {
+
+struct candidateDirection {
+  double cost;
+  double elevation_angle;
+  double azimuth_angle;
+
+  candidateDirection(double c, double e, double z)
+      : cost(c), elevation_angle(e), azimuth_angle(z){};
+
+  bool operator<(const candidateDirection& y) const { return cost < y.cost; }
+
+  bool operator>(const candidateDirection& y) const { return cost > y.cost; }
+};
+}
--- a/local_planner/src/nodes/histogram.cpp
+++ b/local_planner/src/nodes/histogram.cpp
@@ -19,10 +19,8 @@ void Histogram::upsample() {
   resolution_ = resolution_ / 2;
   z_dim_ = 2 * z_dim_;
   e_dim_ = 2 * e_dim_;
-  Eigen::MatrixXi temp_age;
-  Eigen::MatrixXf temp_dist;
-  temp_age.resize(e_dim_, z_dim_);
-  temp_dist.resize(e_dim_, z_dim_);
+  Eigen::MatrixXi temp_age(e_dim_, z_dim_);
+  Eigen::MatrixXf temp_dist(e_dim_, z_dim_);
 
   for (int i = 0; i < e_dim_; ++i) {
     for (int j = 0; j < z_dim_; ++j) {
--- a/local_planner/src/nodes/histogram.cpp
+++ b/local_planner/src/nodes/histogram.cpp
@@ -3,9 +3,8 @@
 
 namespace avoidance {
 Histogram::Histogram(const int res)
-    : resolution_{res}, z_dim_{360 / resolution_}, e_dim_{180 / resolution_} {
-  age_.resize(e_dim_, z_dim_);
-  dist_.resize(e_dim_, z_dim_);
+    : resolution_{res}, z_dim_{360 / resolution_}, e_dim_{180 / resolution_},
+      age_(e_dim_, z_dim_), dist_(e_dim_, z_dim_){
   setZero();
 }
 
@@ -44,10 +43,8 @@ void Histogram::downsample() {
   resolution_ = 2 * resolution_;
   z_dim_ = z_dim_ / 2;
   e_dim_ = e_dim_ / 2;
-  Eigen::MatrixXi temp_age;
-  Eigen::MatrixXf temp_dist;
-  temp_age.resize(e_dim_, z_dim_);
-  temp_dist.resize(e_dim_, z_dim_);
+  Eigen::MatrixXi temp_age(e_dim_, z_dim_);
+  Eigen::MatrixXf temp_dist(e_dim_, z_dim_);
 
   for (int i = 0; i < e_dim_; ++i) {
     for (int j = 0; j < z_dim_; ++j) {
--- a/local_planner/src/nodes/histogram.cpp
+++ b/local_planner/src/nodes/histogram.cpp
@@ -12,9 +12,10 @@ Histogram::Histogram(const int res)
 Histogram::~Histogram() {}
 
 void Histogram::upsample() {
-
-  if(resolution_ != ALPHA_RES * 2){
-	 throw std::logic_error("Invalid use of function upsample(). This function can only be used on a half resolution histogram.");
+  if (resolution_ != ALPHA_RES * 2) {
+    throw std::logic_error(
+        "Invalid use of function upsample(). This function can only be used on "
+        "a half resolution histogram.");
   }
   resolution_ = resolution_ / 2;
   z_dim_ = 2 * z_dim_;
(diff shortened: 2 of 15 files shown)


local_planner/src/nodes/histogram.cpp
```diff
@@ -19,27 +13,19 @@ void Histogram::upsample() {
...
-  std::vector<std::vector<double> > temp_dist;
-  temp_bin.resize(e_dim);
-  temp_age.resize(e_dim);
-  temp_dist.resize(e_dim);
-  for (int i = 0; i < e_dim; ++i) {
-    temp_bin[i].resize(z_dim);
-    temp_age[i].resize(z_dim);
-    temp_dist[i].resize(z_dim);
-  }
+  Eigen::MatrixXd temp_age;
```
nicovanduijn: Can we use floats instead? I don't think we need the precision, and depending on the hardware there might be a significant improvement in performance

nicovanduijn: Also, I'm not familiar with Eigen, but it seems like there should be a way to declare a Matrix with the size right away, avoiding a call to the `resize` method?

local_planner/src/nodes/histogram.cpp
```diff
@@ -48,59 +34,37 @@ void Histogram::downsample() {
...
-        temp_bin[i][j] = 0.0;
-      }
-      temp_age[i][j] = mean_age;
-      temp_dist[i][j] = mean_dist;
+      temp_age(i, j) = mean_age;
+      temp_dist(i, j) = mean_dist;
     }
   }
-  bin = temp_bin;
   age = temp_age;
```
nicovanduijn: are these class members? Then they should be called `age_` and `dist_` instead. On this note, I'd also prefer them to be a bit more descriptive

local_planner/src/nodes/histogram.h
```diff
@@ -31,24 +34,18 @@ class Histogram {
...
+    return dist(x, y);
   }
 
-  inline void set_bin(int x, int y, double value) { bin[x][y] = value; }
-  inline void set_age(int x, int y, double value) { age[x][y] = value; }
-  inline void set_dist(int x, int y, double value) { dist[x][y] = value; }
+  inline void set_age(int x, int y, double value) { age(x, y) = value; }
+  inline void set_dist(int x, int y, double value) { dist(x, y) = value; }
 
   void upsample();
```
nicovanduijn: Maybe a brief comment that explains what `upsample` and `downsample` do. It's non-obvious IMHO, if I understand correctly they always overwrite the current histogram, up/downsample by a a factor of 2. But the user is not allowed to call this function twice. These are crucial assumptions (actually kinda dangerous because they are not enforced anywhere, the code would just segfault if used illegally) so they need to be either well-documented or checked against

local_planner/src/nodes/candidate_direction.h
```diff
@@ -0,0 +1,17 @@
+#pragma once
+
+namespace avoidance {
+
+struct candidateDirection {
+  double cost;
```
mrivi: could be floats right?

local_planner/src/nodes/histogram.cpp
```diff
@@ -1,106 +1,79 @@
...
+
+  if(resolution_ != ALPHA_RES * 2){
+	 throw std::logic_error("Invalid use of function upsample(). This function can only be used on a half resolution histogram.");
   }
-  for (int i = 0; i < e_dim; ++i) {
-    for (int j = 0; j < z_dim; ++j) {
+  resolution_ = resolution_ / 2;
+  z_dim_ = 2 * z_dim_;
+  e_dim_ = 2 * e_dim_;
+  Eigen::MatrixXi temp_age;
```
mrivi: could be a one liner `Eigen::MatrixXi temp_age(e_dim_, z_dim_);`

local_planner/src/nodes/local_planner.cpp
```diff
@@ -273,7 +252,7 @@ void LocalPlanner::determineStrategy() {
              e < goal_e_index + relevance_margin_e_cells; e++) {
           for (int z = goal_z_index - relevance_margin_z_cells;
                z < goal_z_index + relevance_margin_z_cells; z++) {
-            if (polar_histogram_.get_bin(e, z) > 0) {
+            if (polar_histogram_.get_dist(e, z) > 0.001) {
```
mrivi: we could use `FLT_MIN`. There are a couple of these spread trough the PR

local_planner/src/nodes/histogram.cpp
```diff
@@ -1,106 +1,79 @@
...
-    temp_dist[i].resize(z_dim);
+  if(resolution_ != ALPHA_RES){
+	 throw std::logic_error("Invalid use of function downsample(). This function can only be used on a full resolution histogram.");
   }
-  for (int i = 0; i < e_dim; ++i) {
-    for (int j = 0; j < z_dim; ++j) {
+  resolution_ = 2 * resolution_;
+  z_dim_ = z_dim_ / 2;
+  e_dim_ = e_dim_ / 2;
+  Eigen::MatrixXi temp_age;
```
mrivi: same as above

local_planner/src/nodes/planner_functions.cpp
```diff
@@ -301,165 +216,213 @@ void compressHistogramElevation(Histogram& new_hist,
...
+  // pad matrix by smoothing radius respecting all wrapping rules
+  Eigen::MatrixXd matrix_padded;
+  padPolarMatrix(matrix, smoothing_radius, matrix_padded);
+
+  // filter matrix (max-mean)
+  for (int row_index = smoothing_radius;
+       row_index < matrix_padded.rows() - smoothing_radius; row_index++) {
+    for (int col_index = smoothing_radius;
+         col_index < matrix_padded.cols() - smoothing_radius; col_index++) {
+      double original_val = matrix_padded(row_index, col_index);
```
mrivi: float

local_planner/src/nodes/star_planner.cpp
```diff
@@ -157,90 +141,66 @@ void StarPlanner::buildLookAheadTree() {
...
+    std::vector<int> z_FOV_idx;
+    int e_FOV_min, e_FOV_max;
+    calculateFOV(h_FOV_, v_FOV_, z_FOV_idx, e_FOV_min, e_FOV_max,
+                 tree_[origin].yaw_,
+                 0.0);  // assume pitch is zero at every node
+
+    Histogram propagated_histogram = Histogram(2 * ALPHA_RES);
+    Histogram histogram = Histogram(ALPHA_RES);
+
+    propagateHistogram(propagated_histogram, reprojected_points_,
```
mrivi: If I understand correctly `pose_` and `pointcloud_` are not getting updated while being inside the loop ` while (n < n_expanded_nodes_) ` so all the histogram calculations could be done outside the loop not to waste compute.

jkflying:
Based on SITL testing I'm getting occasional tree hits, here's the flight plan I'm using. It is either hitting immediately after takeoff or just before waypoint 5.

[test_mission.zip](https://github.com/PX4/avoidance/files/2841818/test_mission.zip)

local_planner/src/nodes/star_planner.cpp
baumanta: Biiig catch!! But it is the other way round than you thought. The position should of course change. We want to evaluate the possible movement directions at the location of every tree node and not constantly evaluate the current position.

baumanta:
@jkflying with the new tuning I managed to fly the mission you provided in most of the runs. The only critical situation is still before waypoint 5. If the drone steers directly into the hole in the tree, it will see the obstacle only when it is already in a concave gap. But this is a corner case which I think we should look into another time (also I pretty sure master won't be able to handle it either).

baumanta:
I will leave this open until we confirmed SITL results through flight testing

local_planner/src/nodes/histogram.cpp
```diff
@@ -1,106 +1,77 @@
...
-           bin[i_high_res][j_high_res + 1] +
-           bin[i_high_res + 1][j_high_res + 1]) /
-          4.0;
-      double mean_age =
-          (age[i_high_res][j_high_res] + age[i_high_res + 1][j_high_res] +
-           age[i_high_res][j_high_res + 1] +
-           age[i_high_res + 1][j_high_res + 1]) /
-          4.0;
+      float mean_age =
+          (age_(i_high_res, j_high_res) + age_(i_high_res + 1, j_high_res) +
```
jkflying: This can be simplified even further with
`temp_age(i, j) = age_.block(i_high_res, j_high_res, 2, 2).mean();`

I <3 Eigen

baumanta: <3

baumanta:
We flew successful flight tests. Big improvement in terms of safety, a nice distance to the tree is kept at all times. Regression in terms of speed, the drone stopped in front of the tree and took a while to decide which way to go. This problem will be treated in next PR.

PR #197, opened 2019-01-28
Change subscribed node for local velocity

jkflying:
Due to a change in MAVROS 0.28, the local planner with have a bias east in its smoothing function. This is/will be fixed in MAVROS 0.29, however the MAVROS node that needs to be subscribed to has changed.

See https://github.com/mavlink/mavros/pull/1160 and further discussion https://github.com/mavlink/mavros/pull/1142

@TSC21 FYI


jkflying:
Agreed, I like to merge this only once MAVROS 0.29.0 is released anyway.

PR #198, opened 2019-01-30
Waypoint generator tests

mrivi:
Based on  #191 

- test for waypoint generator
- small refactors from double to float
- add comments

### DIFF
--- a/local_planner/src/nodes/avoidance_output.h
+++ b/local_planner/src/nodes/avoidance_output.h
@@ -11,24 +11,36 @@ enum waypoint_choice { hover, costmap, tryPath, direct, reachHeight, goBack };
 
 struct avoidanceOutput {
   waypoint_choice waypoint_type;
+  geometry_msgs::PoseStamped pose;  // current vehicle position
+  bool obstacle_ahead;  // true is there is an obstacle ahead of the vehicle
+  bool reach_altitude;  // true if the vehicle has reach the goal altitude
+  double min_speed;     // minimum speed
+  double max_speed;     // maximum speed
+  double velocity_sigmoid_slope;  // rate at which the speed is linearly
+                                  // increased between min_speed and max_speed
+  ros::Time last_path_time;       // finish built time for the VFH+* tree
 
-  geometry_msgs::PoseStamped pose;
-  bool obstacle_ahead;
-  bool reach_altitude;
-  double min_speed;
-  double max_speed;
-  double velocity_sigmoid_slope;
-  ros::Time last_path_time;
+  geometry_msgs::Point
+      back_off_point;  // closest point to the vehicle in the cloud
+  geometry_msgs::Point back_off_start_point;  // vehicle position when a point
+                                              // in the cloud is closer than
+                                              // min_dist_backoff
+  double min_dist_backoff;  // distance between the vehicle and the closest
+                            // point in the cloud
 
-  geometry_msgs::Point back_off_point;
-  geometry_msgs::Point back_off_start_point;
-  double min_dist_backoff;
+  geometry_msgs::PoseStamped
+      take_off_pose;  // last vehicle position when not armed
+  geometry_msgs::PoseStamped offboard_pose;  // last vehicle position when not
+                                             // in offborad nor in mission mode
 
-  geometry_msgs::PoseStamped take_off_pose;
-  geometry_msgs::PoseStamped offboard_pose;
-
-  double costmap_direction_e;
-  double costmap_direction_z;
-  std::vector<geometry_msgs::Point> path_node_positions;
+  float costmap_direction_e;  // elevation angle of the minimum cost histogram
+                              // cell
+  float
+      costmap_direction_z;  // azimuth angle of the minimum cost histogram cell
+  std::vector<geometry_msgs::Point> path_node_positions;  // array of tree nodes
+                                                          // position, each node
+                                                          // is the minimum cost
+                                                          // node for each tree
+                                                          // depth level
 };
 }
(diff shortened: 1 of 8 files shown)


local_planner/src/nodes/waypoint_generator.h
```diff
@@ -73,36 +73,100 @@ class WaypointGenerator {
...
-  WaypointGenerator();
-  ~WaypointGenerator();
+  /**
+  * @brief     getter method for the system time
+  * @returns   current ROS time
+  **/
+  virtual ros::Time getSystemTime();
+
+  WaypointGenerator() = default;
+  ~WaypointGenerator() = default;
```
jkflying: `virtual ~WaypointGenerator() = default;` since there is a virtual method now.

local_planner/test/test_waypoint_generator.cpp
```diff
@@ -0,0 +1,329 @@
...
+  }
+  void TearDown() override {}
+};
+
+TEST_F(WaypointGeneratorTests, goFastTest) {
+  // GIVEN: a waypoint of type goFast
+
+  // WHEN: we generate waypoints
+  waypointResult result = getWaypoints();
+  result = getWaypoints();
```
jkflying: Why twice? Maybe explain?

local_planner/test/test_waypoint_generator.cpp
```diff
@@ -0,0 +1,330 @@
...
+
+  // THEN: we expect a waypoint in between the vehicle curent position and the
+  // goal
+  EXPECT_FLOAT_EQ(0.9578262852f, result.goto_position.x);
```
jkflying: The values you give here are super precise. It is better to give less precise values for the tests unless the API requires those results explicitly (eg calculating a distance, should be precise). Otherwise we end up testing that the implementation is what we implemented, not that the implementation does what the API says it should.

mrivi:
@jkflying Could you please have a look at commit https://github.com/PX4/avoidance/pull/198/commits/9ac214e6457bbcbd0038c60d8efa91cf0e0dc706 . I changed the logic to only check for progress and not for values. If you're okay with this I'll change the other tests as well.

local_planner/test/test_waypoint_generator.cpp
```diff
@@ -0,0 +1,335 @@
...
+  float pos_sp_to_goal_prev = 1000.0f;
+  double time_sec = 0.33;
+
+  // WHEN: we generate waypoints
+  for (size_t i = 0; i < 5; i++) {
+    waypointResult result = getWaypoints();
+    float goto_to_goal = (toEigen(goal.pose.position) - toEigen(result.goto_position)).norm();
+    float adapted_to_goal = (toEigen(goal.pose.position) - toEigen(result.adapted_goto_position)).norm();
+    float pos_sp_to_goal = (toEigen(goal.pose.position) - toEigen(result.position_waypoint.pose.position)).norm();
+    std::cout << "dist_to_goal " << goto_to_goal << " " << adapted_to_goal << " " << pos_sp_to_goal << std::endl;
```
jkflying: Remove printing

local_planner/src/nodes/avoidance_output.h
```diff
@@ -11,29 +11,36 @@ enum waypoint_choice { hover, costmap, tryPath, direct, reachHeight, goBack };
...
-  geometry_msgs::PoseStamped take_off_pose;
-  geometry_msgs::PoseStamped offboard_pose;
-
-  double costmap_direction_e;
-  double costmap_direction_z;
-  std::vector<geometry_msgs::Point> path_node_positions;
+  geometry_msgs::PoseStamped pose;  // current vehicle position
+  bool obstacle_ahead;  // true is there is an obstacle ahead of the vehicle
+  bool reach_altitude;  // true if the vehicle has reach the goal altitude
+  double min_speed;     // minimum speed
```
nicovanduijn: do we need double precision for these? Same in other places. If the parameters are intrinsically doubles, then sure, but I don't think this is the case?

mrivi: yes, ROS parameters and ros messages like PoseStamped, Point are double. 
From the ROS wiki
- paramtype - defines the type of value stored, and can be any of int_t, double_t, str_t, or bool_t from http://wiki.ros.org/dynamic_reconfigure/Tutorials/HowToWriteYourFirstCfgFile
- 

Primitive Type | Serialization | C++ | Python2 | Python3
-- | -- | -- | -- | --
float64 | 64-bit IEEE float | double | float | float



from http://wiki.ros.org/msg 

I have started converting some double into float here and in #199 little by little.

mrivi:
While rewriting the tests to track the progression towards the goal I came across this incongruency.
https://github.com/PX4/avoidance/blob/master/local_planner/src/nodes/waypoint_generator.cpp#L60
https://github.com/PX4/avoidance/blob/master/local_planner/src/nodes/waypoint_generator.cpp#L40

use the vehicle position from `planner_info_.pose` while in the rest of the code `pose_`. Since the two positions come from different time stamps they are slightly different. I changed everything to `pose_`

local_planner/src/nodes/avoidance_output.h
nicovanduijn: ok, that's what I figured. All good!

jkflying:
From SITL this looks OK. Sometimes I noticed that it seems to  turn in the wrong direction to get to the next waypoint and does a 270 degree turn, maybe there is a quadrant case somewhere wrong? But I don't think that has anything to do with this PR.

mrivi:
@jkflying should we get this in?

PR #199, opened 2019-01-31
Tests star

mrivi:
Based on #195 

- add test for building the tree
- convert variables in `star_planner` from `double` to `float`
- remove unsed

### DIFF
--- a/local_planner/src/nodes/star_planner.h
+++ b/local_planner/src/nodes/star_planner.h
@@ -45,6 +45,8 @@ class StarPlanner {
   geometry_msgs::PoseStamped pose_;
   costParameters cost_params_;
 
+
+
  public:
   std::vector<geometry_msgs::Point> path_node_positions_;
   geometry_msgs::Point obstacle_position_;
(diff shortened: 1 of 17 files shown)


local_planner/test/test_star_planner.cpp
```diff
@@ -0,0 +1,99 @@
...
+TEST_F(StarPlannerTests, buildTree) {
+  // GIVEN: a vehicle position, a goal, and an obstacle in between the stright
+  // line position-goal
+
+  // WHEN: we build the tree
+  star_planner.buildLookAheadTree();
+
+  // THEN: no cell position should be in the obstacle coordinates
+  std::vector<TreeNode> tree_truth;
+  TreeNode n0(0, 0, Eigen::Vector3f(1.2, 0.4, 4));
```
jkflying: Similarly to the other PR, I think we need some other way of making sure this is doing the right thing which doesn't check the exact behavior of the existing implementation, but instead making sure the existing implementation conforms to the API, but will also pass a different implementation conforming to the API. Otherwise we get locked into this implementation, and if we do a rewrite (for whatever reason) the existing tests won't tell us if the new implementation also conforms to the API, but instead that the behavior is identical to the old implementation.

Maybe something like iterating over the points and checking a few things, like that the points move towards the goal, don't get too close to the point cloud, and actually branch outward from the starting pose.

local_planner/src/nodes/star_planner.h
```diff
@@ -45,29 +42,80 @@ class StarPlanner {
   geometry_msgs::PoseStamped pose_;
   costParameters cost_params_;
 
+  /**
```
nicovanduijn: :1st_place_medal:

jkflying:
This looks pretty safe, I'll do a SITL test and if all is good we should be able to merge.

PR #200, opened 2019-01-31
Polar point to histogram index conversion update

AnnaDaiZH:
before elevation angle and azimuth angle were converted separately to histogram index. 
    - the logic for azimuth angle larger than 180 deg and smaller than -180 deg were converted to index 0, which is a valid index.  
   -  the logic for elevation angle larger than 90 deg and smaller than -90 deg were converted to index 0, which is a valid index. 

after change: 
- wrap any elevation and azimuth angle input to valid elevation (-90,90]  and azimuth (-180, 180] angles 
- if the elevation angle was not in the valid angle region before and needs to be wrapped over the top or under neath, the azimuth angle is accordingly wrapped 180 degrees due to the jump
- if the azimuth angle needs to be wrapped, the elevation angle won't be changed as no jump occurred

- depending tests were adapted
- tested in sitl

### DIFF
--- a/local_planner/src/nodes/common.cpp
+++ b/local_planner/src/nodes/common.cpp
@@ -54,27 +54,43 @@ PolarPoint cartesianToPolar(double x, double y, double z,
 Eigen::Vector2i polarToHistogramIndex(const PolarPoint& p_pol, int res) {
   // TODO change logic here, as 0,0 are valid index
   Eigen::Vector2i ev2(0, 0);
-  float e = p_pol.e;
-  float z = p_pol.z;
-  if (res <= 0.f || e < -90.f || e > 90.f || z < -180.f || z > 180.f) {
-    return ev2;
+  PolarPoint p_wrapped = wrapPolar(p_pol);
+  // elevation angle to y-axis histogram index
+  if (p_wrapped.e == 90.0f) {
+    p_wrapped.e = 89.0f;
   }
-  if (e == 90.f) {
-    e = 89;
+  p_wrapped.e += 90.0f;
+  p_wrapped.e = p_wrapped.e + (res - (static_cast<int>(p_wrapped.e) % res));
+  ev2.y() = p_wrapped.e / res - 1;
+  // azimuth angle to x-axis histogram index
+  if (p_wrapped.z == 180.0f) {
+    p_wrapped.z = -180.0f;
   }
-  e += 90.0;
-  e = e + (res - (static_cast<int>(e) % res));  //[-80,+90]
-  ev2.y() = e / res - 1;
-
-  if (z == 180.f) {
-    z = -180;
-  }
-  z += 180.0;
-  z = z + (res - (static_cast<int>(z) % res));  //[-80,+90]
-  ev2.x() = z / res - 1;
+  p_wrapped.z += 180.0f;
+  p_wrapped.z = p_wrapped.z + (res - (static_cast<int>(p_wrapped.z) % res));
+  ev2.x() = p_wrapped.z / res - 1;
   return ev2;
 }
 
+
+PolarPoint wrapPolar(PolarPoint p_pol) {
+  // elevation valid [-90,90)
+  // when abs(elevation) > 90, wrap elevation angle
+  //  azimuth changes for 180 each time
+  while (abs(p_pol.e) > 90.0f) {
+    if (p_pol.e > 90.0f) {
+      p_pol.e = 180.0f - p_pol.e;
+      p_pol.z = p_pol.z - 180.0f;
+    } else if (p_pol.e < -90.0) {
+      p_pol.e = -(180 + p_pol.e);
+      p_pol.z = p_pol.z + 180.0f;
+    }
+  }
+  PolarPoint p_wrapped(p_pol.e, p_pol.z, 0.0f);
+  // azimuth valid [-180,180)
+  wrapAngleToPlusMinus180(p_wrapped.z);
+  return p_wrapped;
+}
 // calculate the yaw for the next waypoint
 double nextYaw(const geometry_msgs::PoseStamped& u,
                const geometry_msgs::Point& v) {
@@ -115,6 +131,16 @@ void wrapAngleToPlusMinusPI(double& angle) {
     angle += 2 * M_PI;
   }
 }
+
+void wrapAngleToPlusMinus180(float& angle) {
+  while (angle > 180.0) {
+    angle -= 360.0;
+  }
+  while (angle < -180.0) {
+    angle += 360.0;
+  }
+}
+
 
 double getAngularVelocity(double desired_yaw, double curr_yaw) {
   wrapAngleToPlusMinusPI(desired_yaw);
--- a/local_planner/src/nodes/common.cpp
+++ b/local_planner/src/nodes/common.cpp
@@ -52,23 +52,21 @@ PolarPoint cartesianToPolar(double x, double y, double z,
 }
 
 Eigen::Vector2i polarToHistogramIndex(const PolarPoint& p_pol, int res) {
-  // TODO change logic here, as 0,0 are valid index
+
   Eigen::Vector2i ev2(0, 0);
   PolarPoint p_wrapped = wrapPolar(p_pol);
-  // elevation angle to y-axis histogram index
+  //elevation angle to y-axis histogram index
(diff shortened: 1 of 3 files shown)


local_planner/src/nodes/common.cpp
```diff
@@ -54,27 +54,42 @@ PolarPoint cartesianToPolar(double x, double y, double z,
 Eigen::Vector2i polarToHistogramIndex(const PolarPoint& p_pol, int res) {
   // TODO change logic here, as 0,0 are valid index
   Eigen::Vector2i ev2(0, 0);
-  float e = p_pol.e;
-  float z = p_pol.z;
-  if (res <= 0.f || e < -90.f || e > 90.f || z < -180.f || z > 180.f) {
-    return ev2;
+  PolarPoint p_wrapped = wrapPolar(p_pol);
+  // elevation angle to y-axis histogram index
+  if (p_wrapped.e == 90.0f) {
```
nicovanduijn: I think this piece of logic doesn't make sense. I believe it was a hack to make things work with integers, I can't see what good it does here. Any opinions @mrivi ?

local_planner/src/nodes/common.cpp
```diff
@@ -54,27 +54,42 @@ PolarPoint cartesianToPolar(double x, double y, double z,
 Eigen::Vector2i polarToHistogramIndex(const PolarPoint& p_pol, int res) {
   // TODO change logic here, as 0,0 are valid index
```
nicovanduijn: remove the TODO comment, since you took care of it

local_planner/src/nodes/common.cpp
nicovanduijn: Actually, after doing some math with pen and paper I see the point now. This is to make sure that the floor() will eventually return a valid index, even if the value is exactly on the upper edge of the last bin. I suggest however, to set it to 89.9f instead, to make it a bit more clear.

local_planner/src/nodes/common.cpp
```diff
@@ -54,27 +54,42 @@ PolarPoint cartesianToPolar(double x, double y, double z,
...
+  PolarPoint p_wrapped = wrapPolar(p_pol);
+  // elevation angle to y-axis histogram index
+  if (p_wrapped.e == 90.0f) {
+    p_wrapped.e = 89.0f;
   }
-  if (e == 90.f) {
-    e = 89;
+  p_wrapped.e += 90.0f;
+  p_wrapped.e = p_wrapped.e + (res - (static_cast<int>(p_wrapped.e) % res));
+  ev2.y() = p_wrapped.e / res - 1;
```
nicovanduijn: These three lines are cryptic and I don't really understand them. What's everyone's confidence in these? I would suggest something like the one-liner: `ev2.y() = floor(p_wrapped.e / res + 90.f / res);`
At the end of the day it's a floor of a linear function that maps elevation -90° to bin 0 and +90° to the highest bin N - 1 (= 180/res -1)

local_planner/src/nodes/common.cpp
```diff
@@ -54,27 +54,42 @@ PolarPoint cartesianToPolar(double x, double y, double z,
...
-  e = e + (res - (static_cast<int>(e) % res));  //[-80,+90]
-  ev2.y() = e / res - 1;
-
-  if (z == 180.f) {
-    z = -180;
-  }
-  z += 180.0;
-  z = z + (res - (static_cast<int>(z) % res));  //[-80,+90]
-  ev2.x() = z / res - 1;
+  p_wrapped.z += 180.0f;
```
nicovanduijn: I suppose we can apply the same logic here. It's a linear function mapping -180° to bin 0 and +180° to bin N -1 = 360/res. Hence: `ev2.x() = floor(p_wrapped.z / res + 180.f / res);`

local_planner/src/nodes/common.cpp
```diff
@@ -54,27 +54,42 @@ PolarPoint cartesianToPolar(double x, double y, double z,
...
+  p_wrapped.z = p_wrapped.z + (res - (static_cast<int>(p_wrapped.z) % res));
+  ev2.x() = p_wrapped.z / res - 1;
   return ev2;
 }
 
+PolarPoint wrapPolar(PolarPoint p_pol) {
```
nicovanduijn: Not the end of the world, but this could be done in-place, similarly to the `wrapAngleToPlusMinus180()`, aka make this `void wrapPolar(PolarPoint& p_pol){....}`

local_planner/src/nodes/common.h
```diff
@@ -73,7 +73,15 @@ PolarPoint cartesianToPolar(double x, double y, double z,
 **/
 
 Eigen::Vector2i polarToHistogramIndex(const PolarPoint& p_pol, int res);
-
+/**
```
nicovanduijn: Love the comments!

local_planner/test/test_common.cpp
```diff
@@ -124,103 +124,60 @@ TEST(Common, elevationAnglefromCartesian) {
...
+  // GIVEN: the polar point and the histogram resolution
   PolarPoint p_pol_1 = {};
   PolarPoint p_pol_2 = {};
   PolarPoint p_pol_3 = {};
   PolarPoint p_pol_4 = {};
   PolarPoint p_pol_5 = {};
-  PolarPoint p_pol_invalid_1 = {};
-  PolarPoint p_pol_invalid_2 = {};
+  PolarPoint p_pol_6 = {};
+  p_pol_1.e = 0.f;
```
nicovanduijn: Use the constructor directly here, will look much nicer. `PolarPoint p_pol_1(34.f, 0.f);` etc...

jkflying:
Just an idea, would it be possible to get correct wrapped versions very easily by converting to cartesian and back? Then we don't have to worry about quadrants etc.

local_planner/src/nodes/common.cpp
AnnaDaiZH: agree, changed to the proposed solution

local_planner/src/nodes/common.cpp
AnnaDaiZH: Yes, changed the mapping to the proposed logic

local_planner/src/nodes/common.cpp
AnnaDaiZH: yes, can be done. now a copy of the original polar point is created in the polarToHistogramIndex, as the input polar point should not be modified in the process.

local_planner/src/nodes/common.cpp
AnnaDaiZH: this case is covered in the support function wrapPolar now.

local_planner/src/nodes/common.cpp
```diff
@@ -52,27 +52,37 @@ PolarPoint cartesianToPolar(double x, double y, double z,
...
+  return ev2;
+}
 
-  if (z == 180.f) {
-    z = -180;
+void wrapPolar(PolarPoint& p_pol) {
+  // elevation valid [-90,90)
+  // when abs(elevation) > 90, wrap elevation angle
+  // azimuth changes for 180 each time
+  while (abs(p_pol.e) > 90.0f) {
```
jkflying: This implementation looks really clean, good work!

local_planner/src/nodes/common.h
```diff
@@ -73,7 +73,14 @@ PolarPoint cartesianToPolar(double x, double y, double z,
...
+* @brief     support function for polarToHistogramIndex
+*            when abs(elevation) > 90, wrap elevation angle into valid
+*            region and azimuth angle changes for +/-180 deg each time
+* @param[in/out]p_pol Polar point with elevation angle [-90,90) and
+*            azimuth angle [-180,180)
+**/
+void wrapPolar(PolarPoint& p_pol);
```
jkflying: There are no tests for wrapPolar directly, I think this would be a pretty easy test to add to help make sure that it works as expected so you don't have to test it in the index conversion tests.

AnnaDaiZH: Agree there should be a new test for this function. I added a test for wrapPolar, but I left the wrapping index tests.

jkflying:
New yaw smoothing fixes the bug with turning in the wrong direction after reaching a waypoint. Also, yaw rates are much smoother.

See this PR before and after the last 2 commits.

image

PR #205, opened 2019-02-07
Yaw smoothing and protect against inf/nan desired setpoints

jkflying:
In case the goal gets set to the exact location of the drone, or at an infinite distance, this protects the yaw from becoming +-Inf or Nan.

It also smooths the yaw so instantaneous setpoint changes don't get propagated to the flight controller.


jkflying:
Successful flight test, the motors were less noisy during major yaw events

PR #208, opened 2019-02-08
Don't check for mission mode when setting trajectory triple points

jkflying:
When the serial communication between FCU and companion is congested, it can result in the first waypoint arriving before the avoidance has switched to mission mode, causing the waypoint to be discarded.

Not checking for mission mode shouldn't have any side effects, since trajectory triples aren't sent unless we want the avoidance running anyway.


nicovanduijn:
Just SITL tested this and it works fine, both with offboard and mission.

LGTM

Issue #179, opened 2019-01-16
Fast take-off in mission mode


nicovanduijn:
During our tests we found that the mission take-off is violently fast when the avoidance is running. This may have gone unnoticed for a long time because we usually take off in position mode before starting the mission.

Sadly, we didn't have logging from boot enabled, so it is not completely captured in the logs below. But we can see that without the avoidance, the beginning of the log still shows the end of a smooth ramp in the z-setpoint, while the log with the enabled OA starts at a 3m altitude (to which it got presumably through a step)

No avoidance:
![bokeh_plot 2](https://user-images.githubusercontent.com/14265408/51246585-09f96a80-198b-11e9-8bf5-33b51d00822d.png)

Avoidance enabled:
![bokeh_plot 1](https://user-images.githubusercontent.com/14265408/51246495-e20a0700-198a-11e9-887d-e22ccce0e51c.png)

mrivi:
@nicovanduijn  I don't know if you are already aware that when OA is running is completely overwrites the auto takeoff coming from the Firmware hence the completely different behaviour
OA auto take off https://github.com/PX4/avoidance/blob/master/local_planner/src/nodes/waypoint_generator.cpp#L202
Firmware auto takeoff https://github.com/PX4/Firmware/blob/master/src/lib/FlightTasks/tasks/AutoMapper2/FlightTaskAutoMapper2.cpp#L135

nicovanduijn:
yes, I'm aware of that. What I'm saying is that the OA's takeoff should be much smoother than it currently is

PR #209, opened 2019-02-11
Smooth takeoff

nicovanduijn:
This commit refactors the waypoint generator in that it uses the smoothing for the take-off scenario as well.

This should address #179.

Flight testing required.

Much better SITL flight performance IMHO
Before: (Notice the step in the setpoint)
bokeh_plot 1

After: Step is gone. It's still not as smooth as it could be, but increasing the smoothness also increases the overshoot, so I settled for this behavior for now. Further tuning could optimize that.
bokeh_plot


local_planner/src/nodes/local_planner_node.cpp
```diff
@@ -470,15 +461,21 @@ void LocalPlannerNode::publishBox() {
 }
 
 void LocalPlannerNode::publishWaypoints(bool hover) {
-  const ros::Time now = ros::Time::now();
+  // Don't publish waypoints if the drone isn't armed or mission has been
+  // enabled. Need second check to avoid race condition
+  if (!armed_ && !mission_) {
```
mrivi: this change breaks offboard and rtl

local_planner/src/nodes/waypoint_generator.cpp
```diff
@@ -191,67 +179,77 @@ bool WaypointGenerator::withinGoalRadius() {
...
+    output_.smoothed_goto_position =
+        pose_.pose.position;  // needs to be local position!
+    return;
+  }
 
-    const Eigen::Vector3f p = location_diff * P_constant;
-    const Eigen::Vector3f d = velocity_diff * D_constant;
+  // If the smoothing speed is set to zero, dont smooth, aka use adapted
+  // waypoint
+  // directly
```
mrivi: could you make it in one line manually please

local_planner/src/nodes/waypoint_generator.cpp
```diff
@@ -191,67 +179,77 @@ bool WaypointGenerator::withinGoalRadius() {
...
-    }
-
-    Eigen::Vector3f velocity_diff =
-        desired_velocity - smoothed_goto_location_velocity_;
-    if (!velocity_diff.allFinite()) {
-      velocity_diff = Eigen::Vector3f::Zero();
-    }
+  // In first iteration (very small or invalid dt), initialize the smoothing
+  // to start at the current drone location
+  if (dt <= 0.01) {
```
mrivi:
```suggestion
  if (dt <= 0.01f) {
```

local_planner/src/nodes/waypoint_generator.cpp
```diff
@@ -191,67 +179,77 @@ bool WaypointGenerator::withinGoalRadius() {
...
+        pose_.pose.position;  // needs to be local position!
+    return;
+  }
 
-    const Eigen::Vector3f p = location_diff * P_constant;
-    const Eigen::Vector3f d = velocity_diff * D_constant;
+  // If the smoothing speed is set to zero, dont smooth, aka use adapted
+  // waypoint
+  // directly
+  if (smoothing_speed_xy_ < 0.01 || smoothing_speed_xy_ < 0.01) {
```
mrivi:
```suggestion
  if (smoothing_speed_xy_ < 0.01f || smoothing_speed_xy_ < 0.01f) {
```

local_planner/src/nodes/waypoint_generator.cpp
```diff
@@ -260,101 +258,49 @@ void WaypointGenerator::nextSmoothYaw(float dt) {
...
+      speed_ * (1.0f - 2 * angle_diff_deg / h_FOV_);  // throttle if outside FOV
+
+  // Scale the pose_to_wp by the speed
   Eigen::Vector3f pose_to_wp =
-      toEigen(output_.adapted_goto_position) - toEigen(pose_.pose.position);
-  if (pose_to_wp.norm() > 0.01f) pose_to_wp.normalize();
-
+      toEigen(output_.goto_position) - toEigen(pose_.pose.position);
+  Eigen::Vector3f pose_to_goal = goal_ - toEigen(pose_.pose.position);
+  if (pose_to_wp.norm() > 0.1) pose_to_wp.normalize();
```
mrivi:
```suggestion
  if (pose_to_wp.norm() > 0.1f) pose_to_wp.normalize();
```

local_planner/src/nodes/waypoint_generator.cpp
```diff
@@ -260,101 +258,49 @@ void WaypointGenerator::nextSmoothYaw(float dt) {
...
   pose_to_wp *= speed_;
 
-  output_.adapted_goto_position =
-      toPoint(toEigen(pose_.pose.position) + pose_to_wp);
+  // If the waypoint is in the direction of the goal, but farther, clamp to goal
+  if (std::atan2(pose_to_wp.cross(pose_to_goal).norm(),
+                 pose_to_wp.dot(pose_to_goal)) < 0.01 &&
```
mrivi:
```suggestion
                 pose_to_wp.dot(pose_to_goal)) < 0.01f &&
```

local_planner/src/nodes/waypoint_generator.cpp
jkflying: Second term should be for `z`, right?

local_planner/src/nodes/waypoint_generator.cpp
```diff
@@ -191,67 +179,77 @@ bool WaypointGenerator::withinGoalRadius() {
...
-    smoothed_goto_location_velocity_ += (p + d) * dt;
-    smoothed_goto_location_ += smoothed_goto_location_velocity_ * dt;
+  // Smooth differently in xz than in z
+  const float P_constant_xy = smoothing_speed_xy_;
+  const float P_constant_z = smoothing_speed_z_;
+  const float D_constant_xy =
+      2.f * std::sqrt(P_constant_xy);  // critically damped
+  const float D_constant_z = 2.f * std::sqrt(P_constant_z);
+  const Eigen::Vector3f P_constant(P_constant_xy, P_constant_xy, P_constant_z);
+  const Eigen::Vector3f D_constant(D_constant_xy, D_constant_xy, D_constant_z);
```
jkflying: Rather calculate D_constant from P_constant with `2*P_constant.array().sqrt()`

Maybe even easier, P_constant and D_constant could be Eigen::Array3f then you wouldn't need all the `array()` calls.

local_planner/src/nodes/local_planner_node.cpp
nicovanduijn: I'm currently experimenting with this part, since it is the cause of some of the issues I'm still having. But why do you think it will break offboard? In offboard the drone is armed, so this clause won't return...

The main problem is that the smoothing starts at the current drone location and then converges toward the setpoint. (Which is the intended behavior I just added in this PR!) But this process starts as soon as the waypoint generator is run the first time, so by the time we actually arm and start, the smoothed setpoint will have converged to the setpoint, hence we're sending a step again. 

I wanted to avoid sending the `armed_` boolean into the waypoint generator (thus further complicating its state machine), but I don't see a way around it.

local_planner/src/nodes/waypoint_generator.cpp
nicovanduijn: done

local_planner/src/nodes/waypoint_generator.cpp
nicovanduijn: done

local_planner/src/nodes/waypoint_generator.cpp
nicovanduijn: good catch. embarrassing

local_planner/src/nodes/waypoint_generator.cpp
nicovanduijn: done

local_planner/src/nodes/waypoint_generator.cpp
nicovanduijn: done

local_planner/src/nodes/waypoint_generator.cpp
nicovanduijn: Good idea. I can actually get rid of the P_constant_xy floats altogether, and this whole block collapses to two lines!
```
  const Eigen::Array3f P_constant(smoothing_speed_xy_, smoothing_speed_xy_, smoothing_speed_z_);
  const Eigen::Array3f D_constant = 2*P_constant.sqrt();
```

local_planner/src/nodes/local_planner_node.cpp
mrivi: yes, sorry I messed up reading the conditions. 
You would like to start generating waypoint when the vehicle is in an OA enabled mode. Would it make sense to check for 
`if(!mission_ && !rtl_ && !offboard_) return;` ?

mrivi: ah no wait.. it could happen in both cases
1. I arm on the ground and stay there for a while
2. I switch to mission mode on the ground and stay there for a while
In both cases you don't want the waypoint generator to start running

mrivi:
```bash
bool in_oa_mode = mission_ || rtl_ || offboard_;
if ( (armed_ && !in_oa_mode) || (!armed && in_oa_mode) ) return;

Comment (nicovanduijn) [reply local_planner/src/nodes/local_planner_node.cpp]: Yeah, I pretty much did that, but passing this boolean (I named it `airborne`) into the waypoint generator anyway. Let me know what you think

File: local_planner/src/nodes/waypoint_generator.cpp [hunk 8]
```diff
@@ -260,101 +260,49 @@ void WaypointGenerator::nextSmoothYaw(float dt) {
...
+      toEigen(output_.goto_position) - toEigen(pose_.pose.position);
+  Eigen::Vector3f pose_to_goal = goal_ - toEigen(pose_.pose.position);
+  if (pose_to_wp.norm() > 0.1f) pose_to_wp.normalize();
   pose_to_wp *= speed_;
 
-  output_.adapted_goto_position =
-      toPoint(toEigen(pose_.pose.position) + pose_to_wp);
+  // If the waypoint is in the direction of the goal, but farther, clamp to goal
+  if (std::atan2(pose_to_wp.cross(pose_to_goal).norm(),
+                 pose_to_wp.dot(pose_to_goal)) < 0.01f &&
```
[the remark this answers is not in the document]
nicovanduijn: I've thought about it, and I guess that's usually safe, as long as we keep path segments short.

It's not completely obvious though, because there could be a path around an obstacle that is in front of the drone (thus dot product of direct-line-to-goal with path is positive) and still has the goal closer than the speed-adapted waypoint.

nicovanduijn:
Flight testing showed lots of issues. One I just fixed with the reset of the smoother (It occurred frequently that the planner reset the smoother because dt was very small) - It seems that sometimes (only on real hardware - this doesn't happen in SITL) the planner runs twice in a row very quickly. Presumably when switching from goStraight() to tryPath() or something like that.

Further issues: 
- the withinGoalRadius() bypassed the smoothing. I took it out completely and it works well in SITL, will flight-test on monday. Since during takeoff the 'goal' is straight up, this lead to a reset of the smoothed position for a brief moment:
newplot

- There seem to be issues when switching from position into mission. If this happens successfully on the first try, all is well. If it goes to hold or switches out of mission and is later enabled again, the smoothed position may be far away... At least that's my theory for now. Will collect more data on monday

nicovanduijn:
Starting to have reasonable performance now. The drops are gone in SITL, and were caused by bad initialization of the smoothing. Flight testing this soon, then hopefully we can merge
bokeh_plot

PR #211, opened 2019-02-11
[WIP] Run LocalPlannerNode at a fixed rate

Jaeyoung-Lim:
This PR moves the main thread of the `LocalPlannerNode` that uses a `std::thread` to use the `ros::AsyncSpinner` and ties it to a timer which runs the node at a fixed rate.

The [AsyncSpinner](http://wiki.ros.org/roscpp/Overview/Callbacks%20and%20Spinning) is a callback function built in for multithreading in roscpp


Jaeyoung-Lim:
After discussion with @baumanta , switched the work queue from the `threadFunction` to replace the while loop.

This allows us to make more objects private within the class

Jaeyoung-Lim:
This has been hard to merge. Will do this after the `main` function becomes more stable.

PR #213, opened 2019-02-12
double/float conversion

mrivi:
Using the compiler flags I have systematically changed `double` into `float` and put casts where needed. 

A few conversions are done in #199 so once that PR is in a will do a final pass on this one.


mrivi:
There is still an issue with polar/cartesian conversion and I am looking into it. Not sure if I screwed up somewhere or having doubles actually makes a difference.

PR #221, opened 2019-02-14
Disable gazebo client by default

nicovanduijn:
Since we have the RVIZ visualization anyway, I disabled the gazebo GUI
by default. This makes SITL run significantly better. Open to discuss
whether we want to merge this or not


jkflying:
I'm 100% for this, the gazebo visualization is missing all the nice annotations you get with RViz, even if it looks a bit prettier.

PR #222, opened 2019-02-14
Realsense exposure tuning

jkflying:
This does 2 fixes:

1. Reduce the average exposure of the realsense, to reduce washout when we are in bright environments. It still does autoexposure, it is just tuned to be darker.

2. Increase the units of the depth image from 1mm to 2mm, so our range increases from 6.5m to 13m.

This needs flight testing before merging, preferably in a bright environment as well as indoors.


jkflying:
Test flight outdoors towards the sun was successful, it saw the tree (just branches, no leaves) much sooner.

Also interesting, it seems the 'indecision' problem was less on this test compared to another vehicle running an earlier firmware.

PR #223, opened 2019-02-14
do not consider histogram cells full if the contain very little points

baumanta:
This got dropped somewhere in all the changes. Originally the was a minimum of points which make up a blocked histogram cell. As it was now, the cell would be blocked even if it contains just one point, which makes the algorithm very sensitive to noise.

Should be flight tested, as I'm not sure, when we lost this and if the chosen default value of 10 points is compatible with the Realsense.

### DIFF
--- a/local_planner/src/nodes/planner_functions.cpp
+++ b/local_planner/src/nodes/planner_functions.cpp
@@ -154,7 +154,7 @@ void propagateHistogram(
 // Generate new histogram from pointcloud
 void generateNewHistogram(Histogram& polar_histogram,
                           const pcl::PointCloud<pcl::PointXYZ>& cropped_cloud,
-                          const Eigen::Vector3f& position) {
+                          const Eigen::Vector3f& position, int n_points_occupied) {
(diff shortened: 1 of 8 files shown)


nicovanduijn:
Is this still relevant after the move to non-binary histogram?

LorenzMeier:
I'm pretty sure it is, since you don't want to respond to basic noise.

local_planner/src/nodes/planner_functions.cpp
```diff
@@ -172,7 +173,7 @@ void generateNewHistogram(Histogram& polar_histogram,
   // Normalize and get mean in distance bins
   for (int e = 0; e < GRID_LENGTH_E; e++) {
     for (int z = 0; z < GRID_LENGTH_Z; z++) {
-      if (counter(e, z) > 0) {
+      if (counter(e, z) > n_points_occupied) {
```
jkflying: This should be `>=`, so that n_points_occupied = 1 works with a single point in the cell.

mrivi:
Superseed by #286

PR #226, opened 2019-02-15
Improve path following behavior. Visualize non-smoothed wp path

baumanta:
In previous flights we sometimes had the issue that at the transition from path following to direct flight weird stuff happened. Sometimes, the waypoint just stayed at the end of the path which made the drone fly a loop.

This is a fix for that. At the end of the path another node is added (same edge length as all others) which points from the last node in the goal direction. This allows for a smoother transition from the path following to direct flight.

The non smoothed waypoint is published as an rviz topic for future debugging

### DIFF
--- a/local_planner/src/nodes/planner_functions.cpp
+++ b/local_planner/src/nodes/planner_functions.cpp
@@ -463,7 +463,10 @@ bool getDirectionFromTree(
                                         path_node_positions.end());
     int size_extended = path_node_positions_extended.size();
 
-    // find path nodes between which the drone is currently located
+    // find path nodes between which the drone is currently located:
+    // a vector with the distances of each node to the drone is generated
+    // the indices of the nodes with smallest and next smallest distance are
+    // found
     int min_dist_idx = 0;
     int second_min_dist_idx = 0;
     float min_dist = HUGE_VAL;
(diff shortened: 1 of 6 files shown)


local_planner/src/nodes/planner_functions.cpp
```diff
@@ -439,23 +439,41 @@ float costFunction(float e_angle, float z_angle, float obstacle_distance,
...
     float second_min_dist = HUGE_VAL;
-    float node_distance =
-        (toEigen(path_node_positions[0]) - toEigen(path_node_positions[1]))
-            .norm();
-
     std::vector<float> distances;
-    distances.reserve(size);
-    for (int i = 0; i < size; i++) {
-      distances.push_back((position - toEigen(path_node_positions[i])).norm());
+    distances.reserve(size_extended);
```
nicovanduijn: I think you can construct it directly with the proper size: `std::vector<float> distances(size_extended);`

local_planner/src/nodes/planner_functions.cpp
```diff
@@ -439,23 +439,41 @@ float costFunction(float e_angle, float z_angle, float obstacle_distance,
...
+
+    for (int i = 0; i < size_extended; i++) {
```
nicovanduijn: you can probably use a range-based for loop here

local_planner/src/nodes/planner_functions.cpp
```diff
@@ -467,15 +485,11 @@ bool getDirectionFromTree(
       }
     }
     int wp_idx = std::min(min_dist_idx, second_min_dist_idx);
-    if (min_dist > 3.0f) {
+
+    if (min_dist > 3.0 ||
```
nicovanduijn: 3.0f

local_planner/test/test_planner_functions.cpp
```diff
@@ -252,11 +252,12 @@ TEST(PlannerFunctions, testDirectionTree) {
   Eigen::Vector3f postion(0.2, 0.3, 1.5);
   Eigen::Vector3f postion1(1.1, 2.3, 2.5);
   Eigen::Vector3f postion2(5.4, 2.0, 2.5);
+  Eigen::Vector3f goal(10, 5, 2.5);
```
nicovanduijn: just to be picky: `10.f, 5.f, 2.5f`

baumanta:
Successfully flight tested on RedLobster (2 missions)

local_planner/src/nodes/planner_functions.cpp
baumanta: I need to access the index to find the closest and second closest node. So I think it would be easiest to leave it that way

PR #228, opened 2019-02-18
fix cost function

baumanta:
this should fix the indecision problem we have seen in SITL as well as in test flights since the cost matrix PR was merged. The drone just could not decide, whether to go left or right and kept hovering and yawing or in the worst case got closer to the obstacle in the process.

### DIFF
--- a/local_planner/src/nodes/local_planner.cpp
+++ b/local_planner/src/nodes/local_planner.cpp
@@ -42,6 +42,7 @@ void LocalPlanner::dynamicReconfigureSetParams(
     avoidance::LocalPlannerNodeConfig &config, uint32_t level) {
   histogram_box_.radius_ = static_cast<float>(config.box_radius_);
   cost_params_.goal_cost_param = config.goal_cost_param_;
+  cost_params_.heading_cost_param = config.heading_cost_param_;
   cost_params_.smooth_cost_param = config.smooth_cost_param_;
   velocity_around_obstacles_ =
       static_cast<float>(config.velocity_around_obstacles_);
@@ -279,13 +280,23 @@ void LocalPlanner::determineStrategy() {
           star_planner_->setReprojectedPoints(reprojected_points_,
                                               reprojected_points_age_);
           star_planner_->setCloud(final_cloud_);
+
+          //set last chosen direction for smoothing
+          PolarPoint last_wp_pol = cartesianToPolar(toEigen(last_sent_waypoint_), toEigen(pose_.pose.position));
+          last_wp_pol.r = (toEigen(pose_.pose.position) - goal_).norm();
+          Eigen::Vector3f projected_last_wp =
+              polarToCartesian(last_wp_pol, pose_.pose.position);
+          star_planner_->setLastDirection(projected_last_wp);
+
+          //build search tree
           star_planner_->buildLookAheadTree();
 
           waypoint_type_ = tryPath;
           last_path_time_ = ros::Time::now();
         } else {
+          float yaw_angle = std::round((-curr_yaw_ * 180.0f / M_PI_F)) + 90.0f;
           getCostMatrix(polar_histogram_, goal_, toEigen(pose_.pose.position),
-                        toEigen(last_sent_waypoint_), cost_params_,
+        		        yaw_angle, toEigen(last_sent_waypoint_), cost_params_,
                         velocity_mod_ < 0.1f, cost_matrix_);
           getBestCandidatesFromCostMatrix(cost_matrix_, 1, candidate_vector_);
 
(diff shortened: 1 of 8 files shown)


jkflying:
Could you explain how taking the distance into account fixes this issue? Was the similarity term in the cost function not working as a result of this?

local_planner/src/nodes/planner_functions.cpp
```diff
@@ -416,7 +420,7 @@ float costFunction(float e_angle, float z_angle, float obstacle_distance,
   // distance cost
   float distance_cost = 0.0f;
   if (obstacle_distance > 0.0f) {
-    distance_cost = 500.0f * 1.0f / obstacle_distance;
+    distance_cost = 1000.0f * 1.0f / obstacle_distance;
```
nicovanduijn: The compiler might optimize this anyway, but why not just `1000.f / obstacle_distance` ?

baumanta:
The cost function was broken after the cost matrix refactor. The points which were compared (candidate, last waypoint, goal) were not projected to the same distance, which had the effect that the smoothing term was so small in comparison to the goal term, that it had no effect. This was fixed.

Also the costfunction did not contain a heading term, which was added. 
In the tree building process the last node position was used as a substitute for the last waypoint, which is incorrect. This does not provide a temporal smoothing effect. This is also fixed by passing the last sent waypoint into the starplanner.

local_planner/test/test_planner_functions.cpp
```diff
@@ -447,6 +447,7 @@ TEST(PlannerFunctions, getCostMatrixNoObstacles) {
   Eigen::Vector3f position(0.f, 0.f, 0.f);
   Eigen::Vector3f goal(0.f, 5.f, 0.f);
   Eigen::Vector3f last_sent_waypoint(0.f, 1.f, 0.f);
+  float heading = 0.f;
```
jkflying: Should another test be added to cover heading changes in the cost function?

baumanta: yes that might be a good idea

baumanta:
We flew very successful flight tests

local_planner/src/nodes/local_planner.cpp
```diff
@@ -279,13 +280,24 @@ void LocalPlanner::determineStrategy() {
           star_planner_->setReprojectedPoints(reprojected_points_,
                                               reprojected_points_age_);
           star_planner_->setCloud(final_cloud_);
+
+          // set last chosen direction for smoothing
+          PolarPoint last_wp_pol = cartesianToPolar(
```
sunzj: The direction vector from last_waypoint to current position should be:
PolarPoint last_wp_pol = cartesianToPolar( position,last_sent_waypoint);

baumanta: I think this line is correct as it is. last_wp_pol is not meant to be a direction vector from last_waypoint to current position. But a polar representation of the last wp in coordinate system with current pos as an origin

sunzj: "a polar representation of the last wp in coordinate system with current pos as an origin" that mean it's a direction vector from current pos ----> last wp. As i understand, if need smoothing the path, it should following last direction: that's last wp ---> current pos.

sunzj: @baumanta the origin is current pos, polar representation of the last wp, it's  the vector from current pos to last wp.

nicovanduijn: Thanks for questioning the logic - that's always healthy! @baumanta and I had quite an in-depth discussion about this as a result actually.

I believe your suggestion does raise a point, but it is not correct per se. last_wp_pol is not meant to be a 'vector', much less one pointing from the last waypoint to the current position. Let me explain:
`last_sent_wp` is not a position we've actually ever reached. Maybe the term "waypoint" is confusing you here - it is merely the "carrot" in the "carrot and stick" analogy that we're playing with the PX4 position controller. The planner runs much faster than it would take the drone to actually travel that far.
image

The way the smoothing works (and this is where I agree with you that it could (and should!) be done better) is that the direction to where the "carrot" was in the previous iteration. This is currently projected out to the goal distance, then this direction is taken into account when calculating the smoothing cost.

Instead, I could imagine that it would be better to take the current motion vector for this, or the polar representation of the previous carrot relative to the previous pose as an approximation thereof.

However, a vector from the previous carrot position to the current drone location is likely to point almost exactly backwards if where we are trying to go.

sunzj: Thanks for your explanation!

I got it, since the planner runs much faster, so the position of  "last_sent_wp" is still a "waypoint" the drone will be reached. Then "last_wp_pol" should be the direction of current position to "last_sent_wp".

And thanks again for your contribution.

baumanta: @sunzj I just opened [PR 252](https://github.com/PX4/avoidance/pull/252) for the changes @nicovanduijn mentioned. Our solution is correct, but maybe not the best one. The PR contains a visualization of the cost terms you might find interesting.

PR #229, opened 2019-02-19
Add common avoidance library as a separate package

Jaeyoung-Lim:
This PR adds a separate package named `avoidance` which hosts common files that can be shared with the global and local planner.

To start with, I have added a companion status publisher in the avoidance library.

By initializing the `CompanionStatusPublisher` object it will automatically publish companion status message without requiring to handle this in the loop of the planner.


Jaeyoung-Lim:
After discussion with @jkflying will close this PR for now

PR #231, opened 2019-02-19
Fix yaml parameters

mrivi:
On master the yaml paramters are not loaded correctly and the default value is always used

### DIFF
--- a/tools/generate_launchfile.sh
+++ b/tools/generate_launchfile.sh
@@ -44,7 +44,7 @@ IFS=";"
 for camera in $CAMERA_CONFIGS; do
 	IFS="," #Inside each camera configuration, the parameters are comma-separated
 	set $camera
-	if [[ $# != 8 ]]; then
+	if [[ $# < 8 ]]; then
 		echo "Invalid camera configuration $camera"
 	else
 		echo "Adding camera $1 with serial number $2"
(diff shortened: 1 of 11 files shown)


jkflying:
No real preference, as long as we are consistent. Having some way for different airframes to have different configurations would be nice too, without requiring different branches of code. Auto-generated yaml files are one option, or something using rqt-reconfigure, but mostly I'd like it to be  consistent.

mrivi:
@nicovanduijn  I can move them to the dynamic reconfigure

nicovanduijn:
I'm not the biggest fan of dynamic reconfigure in terms of making real-world software, but it does come in handy for debugging. So let's go ahead and have all parameters in dynamic reconfigure

mrivi:
@nicovanduijn  @jkflying  How do you feel about having all the parameters in the dynamic reconfigure and load the vehicle specific configuration from a yaml file in the launch script
```bash
<node name="dynparam" pkg="dynamic_reconfigure" type="dynparam" args="load local_planner_node $(find local_planner)/cfg/params_vehicle_1.yaml">
  </node>

**Comment by nicovanduijn** [comment]
I would like that!

**Comment by mrivi** [comment]
Since we are looking at paramters we could transform all these numbers around the codebase into paramters 
https://github.com/PX4/avoidance/blob/master/local_planner/src/nodes/star_planner.cpp#L62

https://github.com/PX4/avoidance/blob/master/local_planner/src/nodes/star_planner.cpp#L65

https://github.com/PX4/avoidance/blob/master/local_planner/src/nodes/star_planner.cpp#L71-L72

**Comment by mrivi** [comment]
@nicovanduijn @jkflying are you good with the changes?

File: tools/generate_launchfile.sh [hunk 1]
```diff
@@ -44,7 +44,7 @@ IFS=";"
 for camera in $CAMERA_CONFIGS; do
 	IFS="," #Inside each camera configuration, the parameters are comma-separated
 	set $camera
-	if [[ $# != 8 ]]; then
+	if [[ $# < 8 ]]; then
```
nicovanduijn: I'm not sure this will work. 

The check for 8 values here is to make sure that each semicolon - separated camera in $CAMERA_CONFIGS has exactly 8 comma-separated values (name, serial number, x, y, z, roll, pitch, yaw). There can be any number of cameras like that.

To pass in an extra argument for yaml file containing the default parameters, I would suggest introducing a new environment variable, since this file is device-specific (and not camera-specific!)

mrivi: @nicovanduijn done

PR #232, opened 2019-02-20
[WIP] Eigen Conersions - Part 2

mrivi:
Converts the remaining part of the code base to Eigen library.
Cherry-pick commits from #226, #209, and #223 to make my life easier when those PRs will be merged into master.

Opens:
- [ ] migrate Quaternion/Euler transformation from Tf package to Eigen. @jkflying any input? is it worth to switch?


jkflying:
Long term I'd like everything not in the node/node_main to have only Eigen types, which would make the algorithms ROS independant, with the node acting as glue for I/O and allowing potentially different frontends (ROS2, something working directly with MAVLINK/RTSP, whatever is created in the future) and thus removing ROS1 as a dependency from the main codebase/algorithms.

Short term, getting whatever we can converted to Eigen will make things consistent and cleaner.

mrivi:
closing in favor of #236

PR #233, opened 2019-02-21
disable yaml cpp build when gazebo simulation is not required

AnnaDaiZH:
add cmake argument -DDISABLE_SIMULATION to disable yaml-cpp build. 

To test the PR
`catkin clean`
`catkin build local_planner -DDISABLE_SIMULATION=ON`
eg. when running the local_planner_sitl_3cam.launch the trees and the drone won't be visible in RViz.

By default, `catkin build local_planner` behaves as before the PR.

### DIFF
--- a/local_planner/CMakeLists.txt
+++ b/local_planner/CMakeLists.txt
@@ -3,6 +3,7 @@ project(local_planner)
 
 add_definitions(-std=c++11)
 
+
 ## Find catkin macros and libraries
 ## if COMPONENTS list like find_package(catkin REQUIRED COMPONENTS xyz)
 ## is used, also find other catkin packages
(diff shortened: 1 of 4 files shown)


local_planner/CMakeLists.txt
```diff
@@ -125,17 +139,32 @@ include_directories(
...
-  src/nodes/histogram.cpp
-  src/nodes/tree_node.cpp
-  src/nodes/box.cpp
-  src/nodes/star_planner.cpp
-  src/nodes/planner_functions.cpp
-  src/nodes/common.cpp
-  src/nodes/rviz_world_loader.cpp
-)
+
+if(DISABLE_SIMULATION)
```
jkflying: Rather than repeating this setup in 2 separate places, add this list of files to a variable, and only if required also append the rviz world loader. Then do the add_library outside of the if.

local_planner/CMakeLists.txt
```diff
@@ -134,9 +150,10 @@ add_library(local_planner
   src/nodes/star_planner.cpp
   src/nodes/planner_functions.cpp
   src/nodes/common.cpp
-  src/nodes/rviz_world_loader.cpp
+  ${RVIZ_WORLD_LOADER}
```
jkflying: Something more like:

```
set(planner_cpp_files a.cpp
                      b.cpp
                      c.cpp)

if ADD_THE_THING
set(planner_cpp_files ${planner_cpp_files}
                      d.cpp)
endif

add_library(local_planner ${planner_cpp_files)

Comment (AnnaDaiZH) [reply local_planner/CMakeLists.txt]: The implemented logic is for not loading the rviz world when the arg is used. Hence the proposed solution is not applicable.

File: local_planner/CMakeLists.txt [hunk 3]
```diff
@@ -125,17 +139,21 @@ include_directories(
...
+set(LOCAL_PLANNER_CPP_FILES   "src/nodes/local_planner.cpp"
+                              "src/nodes/waypoint_generator.cpp"
+                              "src/nodes/histogram.cpp"
+                              "src/nodes/tree_node.cpp"
+                              "src/nodes/box.cpp"
+                              "src/nodes/star_planner.cpp"
+                              "src/nodes/planner_functions.cpp"
+                              "src/nodes/common.cpp"
 )
+if(DISABLE_SIMULATION)
```
jkflying: `if (NOT DISABLE_SIMULATION)`

PR #244, opened 2019-03-07
Test failsafe

mrivi:
- move setting companion status message from `local_planner_node_main` to method in `LocalPlannerNode`
- run roscore in background when launching tests
- test companion status
- rename timeout parameters


jkflying:
I think the valgrind failure is due to the `TransformListener` starting its own thread, but that thread never gets `join()`ed before shutdown. I'm not sure what the best solution is here. Maybe using a tweaked constructor for the test that passes `false` to `spin_thread` in the `TransformListener` constructor?

Also, if there was some way to run these tests without running `roscore` that would be awesome. Maybe we could split the node test into a different test binary, so we don't need to be running `roscore` to run things like the common tests.

mrivi:
I'll split it to a different test binary because I don't think there is a way to run the test without roscore.

.travis.yml
```diff
@@ -73,4 +73,5 @@ script:
   - git config remote.origin.fetch "+refs/heads/*:refs/remotes/origin/*"
   - ./check_code_format.sh
   - catkin build local_planner global_planner --no-deps -v -i --catkin-make-args tests
-  - status=0 && for f in ~/catkin_ws/devel/lib/*/*-test; do valgrind --leak-check=full --track-origins=yes --error-exitcode=1 $f || status=1; done && exit $status
+  - status=0 && for f in ~/catkin_ws/devel/lib/*/*-test; do valgrind --leak-check=full --track-origins=yes --error-exitcode=1 $f || status=1; done
+  - status=0 && (roscore &) && for f in ~/catkin_ws/devel/lib/*/*-test-roscore; do valgrind --leak-check=full --track-origins=yes --error-exitcode=1 $f || status=1; done && exit $status
```
jkflying: This line we don't need the status=0, otherwise we overwrite any errors from the other test run

local_planner/src/nodes/local_planner_node.cpp
```diff
@@ -15,12 +15,16 @@
 
 namespace avoidance {
 
-LocalPlannerNode::LocalPlannerNode() {
+LocalPlannerNode::LocalPlannerNode(const bool tf_spin_thread)
+    : tf_spin_thread_{tf_spin_thread} {
```
jkflying: You could just initialize the `tf_listener_` here, and then you don't need the `tf_spin_thread_` member variable in the class, and the `tf_listener_` doesn't need to be a pointer.

local_planner/include/local_planner/local_planner_node.h
```diff
@@ -101,7 +101,7 @@ enum class MAV_STATE {
 
 class LocalPlannerNode {
  public:
-  LocalPlannerNode();
+  LocalPlannerNode(const bool tf_spin_thread);
```
jkflying: I think it would be cleaner to put ` = true` here to make it the default, then only override it in the test.

local_planner/test/test_local_planner_node.cpp
```diff
@@ -0,0 +1,55 @@
+#include <gtest/gtest.h>
+
+#include "../include/local_planner/local_planner_node.h"
+
+using namespace avoidance;
+
+TEST(LocalPlannerNodeTests, failsafe) {
```
jkflying: This looks great, hopefully in the future we can add more tests here to make sure we are subscribed to and handling inputs correctly.

local_planner/CMakeLists.txt
```diff
@@ -227,12 +227,21 @@ if(CATKIN_ENABLE_TESTING)
 	                                      test/test_planner_functions.cpp
                                         test/test_star_planner.cpp
```
jkflying: The indentation seems funky in here, could you take the opportunity to replace all the tabs with 4 spaces?

jkflying:
SITL tested, all seems good.

baumanta:
hey martina, I just saw that on my computer the tests cannot run anymore. They always get stuck saying:
[ERROR] [1552550422.000052114]: [registerPublisher] Failed to contact master at [localhost:11311].  Retrying...

Do you see the same? Is there a way to fix this?

mrivi:
@baumanta  you need to run roscore to run the local_planner_node tests. Otherwise you can run only the other tests `./devel/lib/local_planner/local_planner-test` and it should be the same as before.

baumanta:
hey @mrivi @jkflying , I just saw that on my computer the tests cannot run anymore. 
If I run:
`cmake --build . --target run_tests`

They always get stuck saying:
[ERROR] [1552550422.000052114]: [registerPublisher] Failed to contact master at [localhost:11311].  Retrying...

Do you see the same? Is there a way to fix this?

PR #246, opened 2019-03-08
remove nans from received sensor message before doing transforms and …

baumanta:
We found out the data from the sensor is heavily padded with nans. In the case of the realsense, the container has a size of about 4'000'000. from which the biggest pointcloud we saw contained 200'000 points. We put all those nans through the transform and into the planner. The planner update function was by far the slowest function in the whole pipeline.

This PR removes the nans before the frame transformation and the update of the planner pointcloud. We see a significant improvement in the execution time of this function even in SITL. We expect this to be even more prominent with the larger realsense pointclouds.

analysis

### DIFF
--- a/local_planner/src/nodes/local_planner_node.cpp
+++ b/local_planner/src/nodes/local_planner_node.cpp
@@ -195,19 +195,18 @@ void LocalPlannerNode::updatePlannerInfo() {
   for (size_t i = 0; i < cameras_.size(); ++i) {
     pcl::PointCloud<pcl::PointXYZ> pcl_cloud_world_frame;
     pcl::PointCloud<pcl::PointXYZ> pcl_cloud_cam_frame;
-    pcl::PointCloud<pcl::PointXYZ> pcl_cloud_cam_frame_without_nan;
     try {
       // transform message to pcl type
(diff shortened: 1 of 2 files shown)


local_planner/src/nodes/local_planner_node.cpp
```diff
@@ -193,18 +193,25 @@ void LocalPlannerNode::updatePlannerInfo() {
...
-                                   cameras_[i].newest_cloud_msg_,
-                                   pc2cloud_world);
-      pcl::fromROSMsg(pc2cloud_world, complete_cloud);
-      local_planner_->complete_cloud_.push_back(std::move(complete_cloud));
+      // transform message to pcl type
+      pcl::fromROSMsg(cameras_[i].newest_cloud_msg_, pcl_cloud_cam_frame);
+
+      // remove nan padding
+      std::vector<int> dummy_index;
+      pcl::removeNaNFromPointCloud(
```
jkflying: According to http://docs.pointclouds.org/trunk/group__filters.html#gac463283a9e9c18a66d3d29b28a575064
The input and output cloud can be the same.

Also reserve dummy_index with the size of the input cloud, to prevent extra allocations.

PR #249, opened 2019-03-12
remove take_off_pose

mrivi:
On master if the planner is killed and then restarted the vehicle cannot advance in the mission because the takeoff position is set in air when the planner is restarted. So `goal.z - 0.5 < take_off_position  + 1 ` . The `starting_height_` can never be reached if `position.z > goal.z` because the waypoint generator will generated setpoints to lower the vehicle altitude.

### DIFF
--- a/local_planner/src/nodes/local_planner.cpp
+++ b/local_planner/src/nodes/local_planner.cpp
@@ -24,7 +24,6 @@ void LocalPlanner::setPose(const Eigen::Vector3f& pos,
   star_planner_->setPose(position_, curr_yaw_histogram_frame_deg_);
 
   if (!currently_armed_ && !disable_rise_to_goal_altitude_) {
-    take_off_pose_ = position_;
     reach_altitude_ = false;
   }
 }
@@ -143,7 +142,7 @@ void LocalPlanner::determineStrategy() {
   if (!reach_altitude_) {
     // use half of the goal altitude as an acceptance altitude as done in
     // the Firmware
-    starting_height_ = std::max(goal_.z() / 2.0f, take_off_pose_.z() + 1.0f);
+    starting_height_ = std::max(goal_.z() / 2.0f, px4_.param_mis_takeoff_alt);
     ROS_INFO("\033[1;35m[OA] Reach height (%f) first: Go fast\n \033[0m",
              starting_height_);
     waypoint_type_ = reachHeight;
@@ -340,7 +339,6 @@ avoidanceOutput LocalPlanner::getAvoidanceOutput() const {
   out.cruise_velocity = px4_.param_mpc_xy_cruise;
   out.last_path_time = last_path_time_;
 
-  out.take_off_pose = take_off_pose_;
   out.starting_height = starting_height_;
 
   out.path_node_positions = star_planner_->path_node_positions_;
--- a/local_planner/src/nodes/local_planner.cpp
+++ b/local_planner/src/nodes/local_planner.cpp
@@ -141,7 +141,9 @@ void LocalPlanner::determineStrategy() {
   }
 
   if (!reach_altitude_) {
-    starting_height_ = std::max(goal_.z() - 0.5f, take_off_pose_.z() + 1.0f);
+    // use half of the goal altitude as an acceptance altitude as done in
+    // the Firmware
+    starting_height_ = std::max(goal_.z() / 2.0f, take_off_pose_.z() + 1.0f);
     ROS_INFO("\033[1;35m[OA] Reach height (%f) first: Go fast\n \033[0m",
              starting_height_);
     waypoint_type_ = reachHeight;
@@ -339,6 +341,7 @@ avoidanceOutput LocalPlanner::getAvoidanceOutput() const {
   out.last_path_time = last_path_time_;
 
   out.take_off_pose = take_off_pose_;
+  out.starting_height = starting_height_;
 
   out.path_node_positions = star_planner_->path_node_positions_;
   return out;
--- a/local_planner/src/nodes/waypoint_generator.cpp
+++ b/local_planner/src/nodes/waypoint_generator.cpp
@@ -131,13 +131,11 @@ void WaypointGenerator::reachGoalAltitudeFirst() {
   // goto_position is a unit vector pointing straight up/down from current
   // location
   output_.goto_position = position_;
-  goal_.x() = position_.x();  // Needed so adaptSpeed can clamp to goal
-  goal_.y() = position_.y();
 
   // Only move the setpoint if drone is in the air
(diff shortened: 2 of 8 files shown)


local_planner/src/nodes/local_planner.cpp
```diff
@@ -167,7 +166,7 @@ void LocalPlanner::determineStrategy() {
   }
 
   if (!reach_altitude_) {
-    starting_height_ = std::max(goal_.z() - 0.5f, take_off_pose_.z() + 1.0f);
```
baumanta: In the case you describe: current position > (goal.z - 0.5) and planner is restarted -> takeoff_pose >(goal.z - 0.5)

Then the starting height will be set to takeoff_pose +1.0 which is higher up than the goal. I agree that this is not ideal but I do not see a critical problem in that. The drone will rise 1m from the current position to the starting height and will start its mission. During the travel to the next wp it will loose the 1m again. Or that is what I would expect to happen. Do you see something different?

mrivi: it can happen that the vehicle never reaches the `starting_height` because `position.z > goal.z` therefore the waypoint generator produces a setpoint to lower the vehicle altitude (https://github.com/PX4/avoidance/blob/master/local_planner/src/nodes/waypoint_generator.cpp#L174-L177).

This is the case I am seeing in simulation.

baumanta: Ah ok , true that makes sense. But then another solution would be to swap the goal.z with starting_height in those lines. Such that it will send waypoints up until the starting height is reached, right? I don't know which is the better solution. I'm just a tiny bit worried that removing the sanity check in starting height will break the offboard workflow.

mrivi: Found another issue: the goal position gets overwritten [here with the current position](https://github.com/PX4/avoidance/blob/master/local_planner/src/nodes/waypoint_generator.cpp#L168-L169).  Since when restarting the node the goal.z and position.z are closer and `speed_` the vehicle is stuck at current position because of https://github.com/PX4/avoidance/blob/master/local_planner/src/nodes/waypoint_generator.cpp#L259-L260

mrivi:
I did the change @baumanta suggested. That lead to finding other issue. When restarting the planner, the vehicle couldn't progress in the mission because the x-y goal was overwritten by the current position and then the setpoint was set to be the goal position. 

I found out that the problem leading to the overshooting at takeoff was due to the fact that the goal position had already switched to the first waypoint while the `starting_height` was not yet reached (planner was still trying to only reach the altitude). This is due to the fact that the Firmware has a much looser condition to set the takeoff point as reached: half the takeoff altitude (https://github.com/PX4/Firmware/blob/master/src/modules/navigator/mission_block.cpp#L198-L214). So  I changed the condition inside the avoidance to match the Firmware. Takeoff performance is still good in simulation. See plots
PR
fix_restart

Master
restart_stuck

local_planner/src/nodes/waypoint_generator.cpp
```diff
@@ -165,13 +165,11 @@ void WaypointGenerator::reachGoalAltitudeFirst() {
...
   // location
   output_.goto_position = position_;
-  goal_.x() = position_.x();  // Needed so adaptSpeed can clamp to goal
-  goal_.y() = position_.y();
 
   // Only move the setpoint if drone is in the air
   if (is_airborne_) {
     // Ascend/Descend to goal altitude
-    if (position_.z() <= goal_.z()) {
+    if (position_.z() <= planner_info_.starting_height) {
```
nicovanduijn: I'm a bit concerned what the `else` here might do. I could imagine the drone just passing the starting height (which is half the goal height), then the goto_position will point straight down until the firmware has updated the next waypoint and thus gotten the planner to switch out of this state...

nicovanduijn:
Is it possible that this somehow bypasses the smoothing now? The take-off looks a bit like a step in that plot, but I haven't actually run a SITL simulation. It's probably just be the time scale of your plot though

local_planner/src/nodes/waypoint_generator.cpp
mrivi: I am not sure I follow

nicovanduijn: Let's say the drone is taking off, `goal_.z` is 4m. Once position_.z is >2.0 m (= starting_height), the goto_position will make the drone descend again, even though we still want to rise

mrivi: If the starting_height has been reached it won't go trought the `reachGoalAltitudeFirst()` because of https://github.com/PX4/avoidance/pull/249/files#diff-2be3b67376d56938a587e393161de500R177

nicovanduijn: Fair enough

jkflying:
Is the half-takeoff altitude relative to the initialization point, or half the difference between the takeoff point and the goal point? If it re-initializes in mid-air then half the altitude will be way down towards the ground.

mrivi:
It's half the goal height of the Firmware mission item

mrivi:
@jkflying @baumanta  I think another solution is
`starting_height = max(goal.z() / 2.f, min_takeoff_alt)` 
where ` min_takeoff_altitude` is the Firmware parameter MIS_TAKEOFF_ALT.  Thoughts?

baumanta:
@mrivi That sounds great! I think using the takeoff altitude from the firmware make sense and also makes all the behavior more consistent. But why would you use goal.z/2 in the first term? I think using the actual goal height would make more sense, no?

mrivi:
@baumanta the problem is that when the the vehicle is still in the state "reachGoalAltitudeFirst", the Firmware has already updated to the next waypoint (The Firmware condition to set the takeoff mission item has reached is half the goal altitude https://github.com/PX4/Firmware/blob/master/src/modules/navigator/mission_block.cpp#L198-L214 ). Therefore the vehicle overshoots when it reaches the takeoff position because the conditions to slow down near the goal are not met. (Before this PR, the problem was solved by overwriting the x-y position of the goal in the waypoint generator when the avoidance was in the state "reachGoalAltitudeFirst")

baumanta:
@mrivi ok, if that is what the firmware does, then I guess we can live with it. Even though I think it is very strange and don't know why the firmware does that in the first place.

mrivi:
Changes done

nicovanduijn:
If you can rebase it, I'll try to test fly it today, so we can get this in. It's been dragging on for too long IMHO

jkflying:
Ok, this probably needs another rebase :sweat_smile: There are some edge cases here that we really need merged.

mrivi:
replaced by #437

PR #250, opened 2019-03-14
Cost matrix image

baumanta:
cost_image
Visualization of the costmatrix, where current heading, chosen waypoint and smoothed waypoint are marked. This 2D visualization is very helpful for debugging

The greener the matrix the more costly from a goal or smoothing perspective the cell is. The reder a cell the more expensive it is due to obstacle distance. The darkest cell is the least expensive one. The dark blue cell (here pink because it is in the red area) is current heading, the light blue one currently chosen best cell, the white one the smoothed setpoint which is actually sent to the pixhawk

Different fixes:
- fix bug in smoothing function
- change smoothing kernel to be non energy conservant
- increase box size
- take out histogram relevance check (get rid of unnecessary moving parts which could cause troubles which are hard to find)

### DIFF
--- a/local_planner/cfg/local_planner_node.cfg
+++ b/local_planner/cfg/local_planner_node.cfg
@@ -7,7 +7,7 @@ gen = ParameterGenerator()
 
 # local_planner
 
-gen.add("box_radius_",    double_t,    0, "Data points farther away will be discarded", 10,  0, 20)
+gen.add("box_radius_",    double_t,    0, "Data points farther away will be discarded", 7,  0, 20)
 gen.add("goal_cost_param_", double_t, 0, "Cost function weight for goal oriented behavior", 3.0, 0.5, 20.0)
 gen.add("heading_cost_param_", double_t, 0, "Cost function weight for constant heading", 0.5, 0.5, 20.0)
 gen.add("smooth_cost_param_", double_t, 0, "Cost function wight for path smoothness", 1.5, 0.5, 20.0)
--- a/local_planner/cfg/local_planner_node.cfg
+++ b/local_planner/cfg/local_planner_node.cfg
@@ -25,6 +25,7 @@ gen.add("reproj_age_", int_t, 0, "maximum age of a reprojected point", 50, 0, 10
 gen.add("velocity_sigmoid_slope_", double_t, 0, "the bigger the bigger the acceleration", 3, 0, 10)
 gen.add("smoothing_speed_xy_", double_t, 0, "response speed of the smoothing system in xy (set to 0 to disable)", 10, 0, 30)
 gen.add("smoothing_speed_z_", double_t, 0, "response speed of the smoothing system in z (set to 0 to disable)", 3, 0, 30)
+gen.add("smoothing_margin_degrees_", double_t, 0, "smoothing radius for obstacle cost in cost histogram", 30, 0, 90)
 
 gen.add("use_vel_setpoints_", bool_t, 0, "Enable velocity setpoints (if false, position setpoints are used)", False)
 gen.add("stop_in_front_", bool_t, 0, "Enable stop in front of the obstacle", False)
--- a/local_planner/cfg/local_planner_node.cfg
+++ b/local_planner/cfg/local_planner_node.cfg
@@ -7,7 +7,7 @@ gen = ParameterGenerator()
 
 # local_planner
 
-gen.add("box_radius_",    double_t,    0, "Minimum box extension in x direction", 7,  0, 20)
+gen.add("box_radius_",    double_t,    0, "Minimum box extension in x direction", 15,  0, 20)
 gen.add("goal_cost_param_", double_t, 0, "Cost function weight for goal oriented behavior", 3.0, 0.5, 20.0)
 gen.add("heading_cost_param_", double_t, 0, "Cost function weight for constant heading", 0.5, 0.5, 20.0)
 gen.add("smooth_cost_param_", double_t, 0, "Cost function wight for path smoothness", 1.5, 0.5, 20.0)
@@ -22,8 +22,6 @@ gen.add("min_dist_backoff_", double_t, 0, "min dist before backing off", 1.5, 0,
 gen.add("timeout_critical_", double_t, 0, "After this timeout the companion status is MAV_STATE_CRITICAL", 0.5, 0, 10)
 gen.add("timeout_termination_", double_t, 0, "After this timeout the companion status is MAV_STATE_FLIGHT_TERMINATION", 15, 0, 1000)
 gen.add("reproj_age_", int_t, 0, "maximum age of a reprojected point", 50, 0, 1000)
-gen.add("relevance_margin_e_degree_", double_t, 0, "obstacles at more than this angle from goal direction are ignored", 25, 0, 90)
-gen.add("relevance_margin_z_degree_", double_t, 0, "obstacles at more than this angle from goal direction are ignored", 40, 0, 180)
 gen.add("velocity_sigmoid_slope_", double_t, 0, "the bigger the bigger the acceleration", 3, 0, 10)
 gen.add("smoothing_speed_xy_", double_t, 0, "response speed of the smoothing system in xy (set to 0 to disable)", 10, 0, 30)
 gen.add("smoothing_speed_z_", double_t, 0, "response speed of the smoothing system in z (set to 0 to disable)", 3, 0, 30)
--- a/local_planner/cfg/local_planner_node.cfg
+++ b/local_planner/cfg/local_planner_node.cfg
@@ -7,7 +7,7 @@ gen = ParameterGenerator()
 
 # local_planner
 
-gen.add("box_radius_",    double_t,    0, "Minimum box extension in x direction", 15,  0, 20)
+gen.add("box_radius_",    double_t,    0, "Data points farther away will be discarded", 10,  0, 20)
 gen.add("goal_cost_param_", double_t, 0, "Cost function weight for goal oriented behavior", 3.0, 0.5, 20.0)
 gen.add("heading_cost_param_", double_t, 0, "Cost function weight for constant heading", 0.5, 0.5, 20.0)
 gen.add("smooth_cost_param_", double_t, 0, "Cost function wight for path smoothness", 1.5, 0.5, 20.0)
--- a/local_planner/src/nodes/common.cpp
+++ b/local_planner/src/nodes/common.cpp
@@ -115,20 +115,20 @@ void createPoseMsg(Eigen::Vector3f& out_waypt, Eigen::Quaternionf& out_q,
           Eigen::AngleAxisf(yaw, Eigen::Vector3f::UnitZ());
 }
 
-float getYawFromQuaternion(const Eigen::Quaternionf q){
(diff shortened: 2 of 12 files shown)


local_planner/cfg/local_planner_node.cfg
```diff
@@ -7,7 +7,7 @@ gen = ParameterGenerator()
 
 # local_planner
 
-gen.add("box_radius_",    double_t,    0, "Minimum box extension in x direction", 7,  0, 20)
+gen.add("box_radius_",    double_t,    0, "Data points farther away will be discarded", 7,  0, 20)
```
nicovanduijn: I said it before and I'll say it again: I don't like the fact that we're violating basic geometry nomenclature.

It's either the side of a cube, the length/width/height of a box or the radius of a sphere...

local_planner/src/nodes/local_planner.cpp
```diff
@@ -216,42 +216,20 @@ void LocalPlanner::determineStrategy() {
...
+      if (hist_is_empty_) {
         obstacle_ = false;
         waypoint_type_ = tryPath;
       }
 
-      if (!hist_is_empty_ && hist_relevant && reach_altitude_) {
+      if (!hist_is_empty_ && reach_altitude_) {
         obstacle_ = true;
 
+        float yaw_angle = std::round((-curr_yaw_ * 180.0f / M_PI_F)) + 90.0f;
```
nicovanduijn: Why do we add 90?

local_planner/src/nodes/local_planner_node.cpp
```diff
@@ -564,8 +564,55 @@ void LocalPlannerNode::publishWaypoints(bool hover) {
   mavros_obstacle_free_path_pub_.publish(obst_free_path);
 }
 
-void LocalPlannerNode::publishHistogramImage() {
+void LocalPlannerNode::publishDataImages() {
+  sensor_msgs::Image cost_img = local_planner_->cost_image_;
+  cost_img.header.stamp = ros::Time::now();
+
+  // current orientation
+  Eigen::Quaternionf q = toEigen(newest_pose_.pose.orientation);
```
nicovanduijn: I believe this whole block appears in many places. I'd suggest a helper function in common 
`float getYawFromPose(const geometry_msgs::PoseStamped& pose)`

local_planner/src/nodes/planner_functions.cpp
```diff
@@ -308,13 +308,56 @@ void getCostMatrix(const Histogram& histogram, const Eigen::Vector3f& goal,
     }
   }
 
-  float smoothing_margin_degrees = 45;
+  float smoothing_margin_degrees = 30;
```
nicovanduijn: either make it a parameter or crystal clear that it is a constant.
`const float SMOOTHING_MARGIN_DEG = 30.0f`

local_planner/src/nodes/planner_functions.cpp
```diff
@@ -308,13 +308,56 @@ void getCostMatrix(const Histogram& histogram, const Eigen::Vector3f& goal,
...
+
+  int image_size = static_cast<int>(image.height * image.width);
+  image.data.clear();
+  image.data.reserve(3 * image_size);
+
+  for (int e = GRID_LENGTH_E - 1; e >= 0; e--) {
+    for (int z = 0; z < GRID_LENGTH_Z; z++) {
+      float distance_cost = 255.f * distance_matrix(e, z) / max_val;
+      float other_cost = 255.f * cost_matrix(e, z) / max_val;
+      image.data.push_back((int)std::max(0.0f, std::min(255.f, distance_cost)));
```
nicovanduijn: `static_cast<int>`

nicovanduijn: Now that I think about it, shouldn't it be a byte? `sensor_msgs::Image` has a uint8...

local_planner/src/nodes/planner_functions.cpp
```diff
@@ -308,13 +308,56 @@ void getCostMatrix(const Histogram& histogram, const Eigen::Vector3f& goal,
...
+      image.data.push_back((int)std::max(0.0f, std::min(255.f, other_cost)));
+
+      image.data.push_back(0);
+    }
+  }
+}
+
+int colorImageIndex(int e_ind, int z_ind, int color) {
+  // color = 0 (red), color = 1 (green), color = 2 (blue)
+  int index = 0;
```
nicovanduijn: how about `return (e_ind * GRID_LENGTH_E + z_ind)*3 + color)`

local_planner/cfg/local_planner_node.cfg
baumanta: the problem is the "box" as it is implemented is neither of all those shapes. In reality it is a sphere with this radius, all points farther away will be ignored. On top of that we use the distance sensor data to determine, which points could belong to the ground and cut off the sphere from the bottom. So I think the description I provided is way more accurate than it was before, but the long explanation just does not fit. We could argue to change the term "box" in the whole code as it is a sphere rather than a box. But actually I just wanted to change the parameter...

local_planner/src/nodes/local_planner.cpp
baumanta: because the histogram angle zero starts at the y-axis, wheras the curr_yaw comes from the firmware and is x-axis oriented

local_planner/src/nodes/planner_functions.cpp
jkflying: It is actually more complicated since e_ind is reversed, so it would be something like

`return ((GRID_LENGTH_E - e_ind * GRID_LENGTH_E) + z_ind) * 3 + color;` , with perhaps an off-by-1 error in there somewhere.

local_planner/src/nodes/planner_functions.cpp
```diff
@@ -498,15 +530,15 @@ void costFunction(float e_angle, float z_angle, float obstacle_distance,
...
+    distance_cost = 700.0f / obstacle_distance;
   }
 
   // combine costs
   other_costs = 0.0f;
-  other_costs =
-      yaw_cost + cost_params.height_change_cost_param_adapted * pitch_cost_up +
-      cost_params.height_change_cost_param * pitch_cost_down + yaw_cost_smooth +
-      pitch_cost_smooth + heading_cost + distance_cost;
+  other_costs = yaw_cost +
```
jkflying: So distance_cost was being added twice before then?

local_planner/src/nodes/star_planner.cpp
```diff
@@ -152,9 +152,11 @@ void StarPlanner::buildLookAheadTree() {
 
     // calculate candidates
     Eigen::MatrixXf cost_matrix;
+    sensor_msgs::Image cost_image;
     std::vector<candidateDirection> candidate_vector;
     getCostMatrix(histogram, goal_, origin_position, tree_[origin].yaw_,
-                  projected_last_wp_, cost_params_, false, cost_matrix);
+                  projected_last_wp_, cost_params_, false, cost_matrix,
+                  cost_image);
```
jkflying: I'd honestly prefer if we could use an `Eigen::MatrixXi` for the cost_image here, so we aren't adding more `sensor_msgs` deep in the algorithm code.

local_planner/src/nodes/local_planner_node.cpp
baumanta: you are right the angle situation with the two conventions is very confusing. I added the function you suggested and tried to clean up a bit at least such that the variable names reflect the coordinate system.

local_planner/src/nodes/planner_functions.cpp
baumanta: made it a parameter

local_planner/src/nodes/common.cpp
```diff
@@ -115,6 +115,22 @@ void createPoseMsg(Eigen::Vector3f& out_waypt, Eigen::Quaternionf& out_q,
           Eigen::AngleAxisf(yaw, Eigen::Vector3f::UnitZ());
 }
 
+float getYawFromQuaternion(const Eigen::Quaternionf q){
```
nicovanduijn: From wikipedia: 
```
	double siny_cosp = +2.0 * (q.w() * q.z() + q.x() * q.y());
	double cosy_cosp = +1.0 - 2.0 * (q.y() * q.y() + q.z() * q.z());  
	yaw = atan2(siny_cosp, cosy_cosp);
```

local_planner/src/nodes/common.cpp
```diff
@@ -115,6 +115,22 @@ void createPoseMsg(Eigen::Vector3f& out_waypt, Eigen::Quaternionf& out_q,
...
+	tf::Quaternion tf_q(q.x(), q.y(), q.z(), q.w());
+	tf::Matrix3x3 tf_m(tf_q);
+	double roll, pitch, yaw;
+	tf_m.getRPY(roll, pitch, yaw);
+	return static_cast<float>(yaw);
+}
+
+float getPitchFromQuaternion(const Eigen::Quaternionf q){
+	tf::Quaternion tf_q(q.x(), q.y(), q.z(), q.w());
```
nicovanduijn: Again, from wiki: (https://en.wikipedia.org/wiki/Conversion_between_quaternions_and_Euler_angles)
(Maybe we'll have to verify the frame and convention first, but I'd prefer doing this in two simple lines as opposed to calling the tf libraries)
```
	double sinp = +2.0 * (q.w() * q.y() - q.z() * q.x());
	if (fabs(sinp) >= 1)
		pitch = copysign(M_PI / 2, sinp); // use 90 degrees if out of range
	else
		pitch = asin(sinp);
```

local_planner/src/nodes/planner_functions.cpp
baumanta: and the right answer iiiiiiiis:   ((GRID_LENGTH_E - e_ind - 1) * GRID_LENGTH_Z + z_ind) * 3 + color;

local_planner/src/nodes/planner_functions.cpp
baumanta: yes, this was a bug

local_planner/src/nodes/common.cpp
baumanta: I prefer doing that in a seperate PR. There are so many conventions for Quaternion and it took me some time to find this one (which actually works). I agree that we should find a more elegant way. But I think we also agreed to try and keep PRs as small and testable as possible

local_planner/src/nodes/local_planner.cpp
```diff
@@ -125,38 +125,30 @@ void LocalPlanner::create2DObstacleRepresentation(const bool send_to_fcu) {
...
-  image.encoding = sensor_msgs::image_encodings::MONO8;
-  image.is_bigendian = 0;
-  image.step = 255;
-
-  int image_size = static_cast<int>(image.height * image.width);
-  image.data.reserve(image_size);
+void LocalPlanner::generateHistogramImage(Histogram &histogram) {
+  histogram_image_data_.clear();
+  histogram_image_data_.reserve(GRID_LENGTH_E * GRID_LENGTH_Z);
+  float sensor_max_dist = 10.f;
```
mhkabir: Why?

baumanta:
We would like to to one more test flight, then this should be ready to go. Praying for nicer weather...

PR #252, opened 2019-03-15
save last waypoint for temporal smoothing in agular representation

baumanta:
The last waypoint was so far saved as a cartesion point. That meant the smoothing encouraged going towards this point absolute in space.

I would suggest a better solution would be to safe the direction we were commanding the vehicle to go in the last timestep. Smoothing would then encourage going into the same direction. This visualization shows the difference. The colored points goal, proj_wp, proj_heading are used to determine the cost of the proj_candidate in the cost function.

This PR would need some SITL testing and flight testing.
final

### DIFF
--- a/local_planner/include/local_planner/local_planner.h
+++ b/local_planner/include/local_planner/local_planner.h
@@ -5,6 +5,7 @@
 #include "avoidance_output.h"
 #include "box.h"
 #include "candidate_direction.h"
+#include "common.h"
 #include "cost_parameters.h"
 #include "histogram.h"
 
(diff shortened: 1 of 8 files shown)


jkflying:
Interesting, this definitely needs SITL testing and maybe a faster smoothing to compensate, since it now pulls the setpoint along with the drone.

baumanta:
I don't think this is a necessary fix. This PR was the result of a discussion with @nicovanduijn. But it is not clear if the change would improve or worsen the current state. Therefore closing for now

PR #253, opened 2019-03-15
Add link time optimization for the library and node

jkflying:
This adds Link Time Optimization when compiling in Release mode. It should improve speed in the helper functions for things like polarToCartesian which are called ~500k times per iteration, and reduce binary size considerably. Note that it does slow down linking.

Release build sizes before:
devel/lib/local_planner/local_planner_node 103K
devel/lib/liblocal_planner.so 1.1M 

After:
devel/lib/local_planner/local_planner_node 78K
devel/lib/liblocal_planner.so 878K

Then adding static linking:
devel/lib/local_planner/local_planner_node 762K

### DIFF
--- a/local_planner/CMakeLists.txt
+++ b/local_planner/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8.3)
+cmake_minimum_required(VERSION 3.9)
 project(local_planner)
 
 add_definitions(-std=c++11)
@@ -164,6 +164,18 @@ add_dependencies(local_planner ${${PROJECT_NAME}_EXPORTED_TARGETS} ${catkin_EXPO
 ## Declare a C++ executable
 # add_executable(avoidance_node src/avoidance_node.cpp)
 add_executable(local_planner_node src/nodes/local_planner_node_main.cpp)
+
+## Check for Link Time Optimization support
+## and enable it on the library and node if we are doing a release build
+include(CheckIPOSupported)
+check_ipo_supported(RESULT ipo_supported OUTPUT error)
+if( ipo_supported AND ${CMAKE_BUILD_TYPE} STREQUAL "Release")
+    message(STATUS "IPO / LTO enabled")
+    set_property(TARGET local_planner PROPERTY INTERPROCEDURAL_OPTIMIZATION TRUE)
+    set_property(TARGET local_planner_node PROPERTY INTERPROCEDURAL_OPTIMIZATION TRUE)
+else()
+    message(STATUS "IPO / LTO not enabled: <${error}>")
+endif()
 
 ## Add cmake target dependencies of the executable
 ## same as for the library above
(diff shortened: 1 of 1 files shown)


jkflying:
@dagar Has there been any investigation of using LTO on the firmware? I took a quick look and didn't see any references, maybe I'm missing something.

dagar:
> @dagar Has there been any investigation of using LTO on the firmware? I took a quick look and didn't see any references, maybe I'm missing something.

It's something I've wanted to do for a long time, but kept running into issues requiring non-trivial code changes. I'll likely take another stab at it the next time we bump the toolchain. In the meantime we have a poor man's version on a per module basis (https://github.com/PX4/Firmware/blob/b54ca67de9b5900c7ed11bef323c7b4e5a0339b6/src/modules/mavlink/CMakeLists.txt#L71).

Somewhat relatedly, the avoidance project might be a good candidate for profile guided optimization (PGO) where you build with coverage, run a representative workload, then build again with that added information. It adds a fair amount of build complexity, but can yield significant improvements. It can also be an optional step mainly done for packages.

jkflying:
> It's something I've wanted to do for a long time, but kept running into issues requiring non-trivial code changes. I'll likely take another stab at it the next time we bump the toolchain. In the meantime we have a poor man's version on a per module basis (https://github.com/PX4/Firmware/blob/b54ca67de9b5900c7ed11bef323c7b4e5a0339b6/src/modules/mavlink/CMakeLists.txt#L71).

Interesting! I did a quick `grep` of the Firmware repo and it seems `mavlink` is the only module using it - there might be some reasonable gains waiting there, particularly for things like matrix libraries where dependency analysis can show that not all of the returned elements are used.

> Somewhat relatedly, the avoidance project might be a good candidate for profile guided optimization (PGO) where you build with coverage, run a representative workload, then build again with that added information. It adds a fair amount of build complexity, but can yield significant improvements. It can also be an optional step mainly done for packages.

Agreed, but I think there are still easier wins for performance optimization at an algorithmic level before we resort to PGO. PGO is where we'll look once we're out of ideas ;-)

dagar:
> > It's something I've wanted to do for a long time, but kept running into issues requiring non-trivial code changes. I'll likely take another stab at it the next time we bump the toolchain. In the meantime we have a poor man's version on a per module basis (https://github.com/PX4/Firmware/blob/b54ca67de9b5900c7ed11bef323c7b4e5a0339b6/src/modules/mavlink/CMakeLists.txt#L71).
> 
> Interesting! I did a quick `grep` of the Firmware repo and it seems `mavlink` is the only module using it - there might be some reasonable gains waiting there, particularly for things like matrix libraries where dependency analysis can show that not all of the returned elements are used.

It's mildly annoying for debugging or even compile errors that now reference a giant temporary single source file.

jkflying:
This is blocked by an old version of CMake by default in 16.04. This can wait until 18.04 is the main target for PX4/Avoidance for merging.

PR #254, opened 2019-03-18
Quaternion angle cleanup

baumanta:
There has long been a bit of a chaos with the angle names and frames: The histogram yaw angle is 90 deg shifted from the fcu yaw and measured in the other direction. The variables were not properly named which was dangerous, also the variable names did not reflect whether an angle was radian or degree. I tried to name them properly to avoid future confusion.

In the process I found a major bug in `calculateFOV` sometimes the input angles were given in radian and sometimes in degrees. I fixed this function, also cleaning it up and using the simple logic now provided in common.

### DIFF
--- a/local_planner/include/local_planner/local_planner.h
+++ b/local_planner/include/local_planner/local_planner.h
@@ -57,8 +57,9 @@ class LocalPlanner {
   int reproj_age_;
   int counter_close_points_backoff_ = 0;
 
-  float curr_yaw_deg_fcu_frame_, curr_yaw_deg_histogram_frame_;
-  float curr_pitch_deg_;
+  float curr_yaw_fcu_frame_deg_, curr_yaw_histogram_frame_deg_;
+  float curr_pitch_deg_;  // for pitch angles the histogram frame matches the
+                          // fcu frame
   float velocity_around_obstacles_;
   float velocity_far_from_obstacles_;
   float keep_distance_;
--- a/local_planner/include/local_planner/local_planner.h
+++ b/local_planner/include/local_planner/local_planner.h
@@ -57,8 +57,8 @@ class LocalPlanner {
   int reproj_age_;
   int counter_close_points_backoff_ = 0;
 
-  float curr_yaw_fcu_frame_;
-  float curr_pitch_fcu_frame_;
+  float curr_yaw_deg_fcu_frame_, curr_yaw_deg_histogram_frame_;
+  float curr_pitch_deg_;
   float velocity_around_obstacles_;
   float velocity_far_from_obstacles_;
   float keep_distance_;
--- a/local_planner/include/local_planner/planner_functions.h
+++ b/local_planner/include/local_planner/planner_functions.h
@@ -65,7 +65,7 @@ void calculateFOV(float h_FOV, float v_FOV, std::vector<int>& z_FOV_idx,
 * @param[in] reprojected_points, previous iterations occupied histogram bins
 *reprojected in 3D space
 * @param[in] reprojected_points_age, age of each reprojected point
-* @param[]   position, current vehicle positon
+* @param[in]   position, current vehicle positon
 **/
 void propagateHistogram(
     Histogram& polar_histogram_est,
@@ -164,13 +164,14 @@ void getBestCandidatesFromCostMatrix(
 
 /**
 * @brief   computes the cost of each direction in the polar histogram
-* @param[] e_angle, elevation angle [deg]
-* @param[] z_angle, azimuth angle [deg]
-* @param[] goal, current goal position
-* @param[] position, current vehicle position
-* @param[] position, current vehicle heading in histogram angle convention [deg]
-* @param[] last_sent_waypoint, previous position waypoint
-* @param[] cost_params, weights for goal oriented vs smooth behaviour
+* @param[in] e_angle, elevation angle [deg]
+* @param[in] z_angle, azimuth angle [deg]
+* @param[in] goal, current goal position
+* @param[in] position, current vehicle position
+* @param[in] position, current vehicle heading in histogram angle convention
+*[deg]
+* @param[in] last_sent_waypoint, previous position waypoint
+* @param[in] cost_params, weights for goal oriented vs smooth behaviour
 * @param[out] distance_cost, cost component due to proximity to obstacles
 * @param[out] other_costs, cost component due to goal and smoothness
 **/
@@ -184,7 +185,7 @@ void costFunction(float e_angle, float z_angle, float obstacle_distance,
 /**
 * @brief   max-median filtes the cost matrix
 * @param   matrix, cost matrix
-* @param[] smoothing_radius, median filter window size
+* @param[in] smoothing_radius, median filter window size
 **/
 void smoothPolarMatrix(Eigen::MatrixXf& matrix, unsigned int smoothing_radius);
 
@@ -208,7 +209,7 @@ Eigen::ArrayXf getConicKernel(int radius);
 
 /**
 * @brief   helper method to output on the console the histogram
-* @param[] histogram, polar histogram
+* @param[in] histogram, polar histogram
 **/
 void printHistogram(Histogram& histogram);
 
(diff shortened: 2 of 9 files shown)


local_planner/include/local_planner/local_planner.h
```diff
@@ -57,8 +57,8 @@ class LocalPlanner {
   int reproj_age_;
   int counter_close_points_backoff_ = 0;
 
-  float curr_yaw_fcu_frame_;
-  float curr_pitch_fcu_frame_;
+  float curr_yaw_deg_fcu_frame_, curr_yaw_deg_histogram_frame_;
```
nicovanduijn: Purely a matter of taste, but I find it a big strange that pitch doesn't specify the frame while yaw does. (I know that there is no difference between the two frames for pitch, but not everyone might know that). Also, I prefer the unit to be specified last, so it is more consistent with naming of things that don't have a specified frame (e.g. `back_off_dist_m_` or something like that). With time, you'll get used to variables having the unit specified in the end

local_planner/test/test_local_planner.cpp
```diff
@@ -87,7 +87,7 @@ TEST_F(LocalPlannerTests, all_obstacles) {
 
   for (size_t i = 0; i < scan.ranges.size(); i++) {
     // width determined empirically, TODO investigate why it isn't symmetrical
```
nicovanduijn: Does that take care of this TODO?

baumanta: Julian wrote the TODO and I'm not completely sure if that fixes what he meant.

nicovanduijn: @jkflying

local_planner/include/local_planner/planner_functions.h
```diff
@@ -168,15 +168,15 @@ void getBestCandidatesFromCostMatrix(
 * @param[] z_angle, azimuth angle [deg]
 * @param[] goal, current goal position
 * @param[] position, current vehicle position
-* @param[] position, current vehicle heading in histogram angle convention
+* @param[] position, current vehicle heading in histogram angle convention [deg]
```
nicovanduijn: `position` -> `yaw_angle_histogram_frame_deg`
`@param[]` -> `@param[in]`

local_planner/include/local_planner/star_planner.h
```diff
@@ -28,7 +28,7 @@ class StarPlanner {
   float tree_node_distance_ = 1.0f;
   float tree_discount_factor_ = 0.8f;
   float max_path_length_ = 4.f;
-  float curr_yaw_deg_histogram_frame_;
+  float curr_yaw_histogram_frame_deg_;
```
nicovanduijn: uninitialized

local_planner/src/nodes/common.cpp
```diff
@@ -115,22 +115,22 @@ void createPoseMsg(Eigen::Vector3f& out_waypt, Eigen::Quaternionf& out_q,
 }
 
 float getYawFromQuaternion(const Eigen::Quaternionf q) {
-  float siny_cosp = +2.0 * (q.w() * q.z() + q.x() * q.y());
-  float cosy_cosp = +1.0 - 2.0 * (q.y() * q.y() + q.z() * q.z());
+  float siny_cosp = 2.0 * (q.w() * q.z() + q.x() * q.y());
```
nicovanduijn: 2.0f

local_planner/src/nodes/common.cpp
```diff
@@ -115,22 +115,22 @@ void createPoseMsg(Eigen::Vector3f& out_waypt, Eigen::Quaternionf& out_q,
...
+  float cosy_cosp = 1.0 - 2.0 * (q.y() * q.y() + q.z() * q.z());
   float yaw = atan2(siny_cosp, cosy_cosp);
 
-  return static_cast<float>(yaw);
+  return static_cast<float>(yaw * RAD_TO_DEG);
 }
 
 float getPitchFromQuaternion(const Eigen::Quaternionf q) {
   float pitch = 0;
```
nicovanduijn: 0.0f

local_planner/src/nodes/common.cpp
```diff
@@ -115,22 +115,22 @@ void createPoseMsg(Eigen::Vector3f& out_waypt, Eigen::Quaternionf& out_q,
...
   float sinp = 2.0 * (q.w() * q.y() - q.z() * q.x());
   if (fabs(sinp) >= 1) {
```
nicovanduijn: 1.0f

local_planner/src/nodes/common.cpp
```diff
@@ -115,22 +115,22 @@ void createPoseMsg(Eigen::Vector3f& out_waypt, Eigen::Quaternionf& out_q,
...
 
 float getYawFromQuaternion(const Eigen::Quaternionf q) {
-  float siny_cosp = +2.0 * (q.w() * q.z() + q.x() * q.y());
-  float cosy_cosp = +1.0 - 2.0 * (q.y() * q.y() + q.z() * q.z());
+  float siny_cosp = 2.0 * (q.w() * q.z() + q.x() * q.y());
+  float cosy_cosp = 1.0 - 2.0 * (q.y() * q.y() + q.z() * q.z());
   float yaw = atan2(siny_cosp, cosy_cosp);
 
-  return static_cast<float>(yaw);
+  return static_cast<float>(yaw * RAD_TO_DEG);
```
nicovanduijn: Is this cast needed? I guess `RAD_TO_DEG` is a float already, so you can remove it

local_planner/src/nodes/common.cpp
```diff
@@ -115,22 +115,22 @@ void createPoseMsg(Eigen::Vector3f& out_waypt, Eigen::Quaternionf& out_q,
...
 }
 
 float getPitchFromQuaternion(const Eigen::Quaternionf q) {
   float pitch = 0;
   float sinp = 2.0 * (q.w() * q.y() - q.z() * q.x());
   if (fabs(sinp) >= 1) {
-    pitch = copysign(M_PI / 2, sinp);  // use 90 degrees if out of range
+    pitch = copysign(M_PI / 2, sinp);  // use PI/2 if out of range
```
nicovanduijn: 2.0f

local_planner/src/nodes/common.cpp
```diff
@@ -115,22 +115,22 @@ void createPoseMsg(Eigen::Vector3f& out_waypt, Eigen::Quaternionf& out_q,
...
   } else {
     pitch = asin(sinp);
   }
-  return static_cast<float>(pitch);
+  return static_cast<float>(pitch * RAD_TO_DEG);
```
nicovanduijn: Again, I think you can remove this cast as well

local_planner/src/nodes/common.cpp
```diff
@@ -115,22 +115,22 @@ void createPoseMsg(Eigen::Vector3f& out_waypt, Eigen::Quaternionf& out_q,
...
+  float cosy_cosp = 1.0 - 2.0 * (q.y() * q.y() + q.z() * q.z());
   float yaw = atan2(siny_cosp, cosy_cosp);
 
-  return static_cast<float>(yaw);
+  return static_cast<float>(yaw * RAD_TO_DEG);
 }
 
 float getPitchFromQuaternion(const Eigen::Quaternionf q) {
   float pitch = 0;
   float sinp = 2.0 * (q.w() * q.y() - q.z() * q.x());
```
nicovanduijn: 2.0f

local_planner/src/nodes/common.cpp
```diff
@@ -115,22 +115,22 @@ void createPoseMsg(Eigen::Vector3f& out_waypt, Eigen::Quaternionf& out_q,
 }
 
 float getYawFromQuaternion(const Eigen::Quaternionf q) {
-  float siny_cosp = +2.0 * (q.w() * q.z() + q.x() * q.y());
-  float cosy_cosp = +1.0 - 2.0 * (q.y() * q.y() + q.z() * q.z());
+  float siny_cosp = 2.0 * (q.w() * q.z() + q.x() * q.y());
+  float cosy_cosp = 1.0 - 2.0 * (q.y() * q.y() + q.z() * q.z());
```
nicovanduijn: same

local_planner/test/test_local_planner.cpp
jkflying: Are they symmetrical now? We should be able to determine the indexes based on the geometry to the obstacles we generate.

baumanta:
In SITL it seemed to work fine with the current cost function. After all the bug did not affect the first tree node (current position) but only all the deeper ones. But we should definitely flight test it.

PR #255, opened 2019-03-19
Visualization outfactor

baumanta:
To clean up the code a bit and make it fit for future use (maybe use different planners at some point).

I took out all the RVIZ markers ant put them into a separate class. Now the node only contains the topics uses for navigation (to/from mavros, realsense).


local_planner/include/local_planner/local_planner_node.h
```diff
@@ -214,17 +203,18 @@ class LocalPlannerNode {
   void fillUnusedTrajectoryPoint(mavros_msgs::PositionTarget& point);
 
   /**
-  * @brief     publishes position and velocity setpoints both to the FCU and to
-  *Rviz for visualization
+  * @brief     calculateds position and velocity setpoints and sends to the FCU
```
nicovanduijn: - `calculates`
- remove `and to RViiz for visualization`
- Fix awkward line break here

local_planner/include/local_planner/local_planner_visualization.h
```diff
@@ -0,0 +1,164 @@
...
+#include <vector>
+
+namespace avoidance {
+
+class LocalPlannerVisualization {
+ public:
+  /**
+  * @brief      initializes all subscribers used for local planner visualization
+  **/
+  void initializeSubscribers(ros::NodeHandle& nh);
```
nicovanduijn: publishers!

local_planner/include/local_planner/local_planner_visualization.h
```diff
@@ -0,0 +1,164 @@
...
+  /**
+  * @brief       Main function which calls functions to visualize all planner
+  *output ready at the end of one planner iteration
+  * @params[in] planner, reference to the planner
+  * @params[in] newest_waypoint_position, last caluclated waypoint (smoothed)
+  * @params[in] newest_adapted_waypoint_position, last caluclated waypoint
+  *(non-smoothed)
+  **/
+  void visualizePlannerData(
+      LocalPlanner& planner,
```
nicovanduijn: This is somewhat questionable - If the LocalPlanner owns a LocalPlannerVisualization, which has a reference to a LocalPlanner, you create a cyclic dependency. It's possible - but not really good OO architecture I think

local_planner/include/local_planner/local_planner_visualization.h
```diff
@@ -0,0 +1,164 @@
...
+  /**
+  * @brief       Visualization of the calculated search tree and the best path
+  *chosen
+  * @params[in]  tree, the complete calculated search tree
+  * @params[in]  closed_set, the closed set (all expanded nodes)
+  * @params[in]  path_node_positions, the positions of all nodes belonging to
+  *the chosen best path
+  **/
+  void publishTree(std::vector<TreeNode>& tree,
+                   const std::vector<int> closed_set,
```
nicovanduijn: There's no point in passing a const value.
Either const& or value (here and all following functions too)

local_planner/src/nodes/local_planner_node.cpp
```diff
@@ -64,33 +64,10 @@ LocalPlannerNode::LocalPlannerNode(const bool tf_spin_thread) {
   px4_param_sub_ = nh_.subscribe("/mavros/param/param_value", 1,
                                  &LocalPlannerNode::px4ParamsCallback, this);
 
+  // initialize standard publishers
   world_pub_ = nh_.advertise<visualization_msgs::MarkerArray>("/world", 1);
   drone_pub_ = nh_.advertise<visualization_msgs::Marker>("/drone", 1);
```
nicovanduijn: Why are these two not moved into the visualizer?

local_planner/include/local_planner/local_planner_visualization.h
jkflying: Agreed, circular dependencies are bad. Either the local planner should pass the data into the visualization (maybe with a struct?), or the local planner shouldn't own the copy of the visualization (maybe it lives in the node instead.

local_planner/include/local_planner/local_planner_visualization.h
jkflying: Yep, I think const& was intended here, and a few other places too.

local_planner/include/local_planner/local_planner_visualization.h
baumanta: In my opinion this is not a cyclic dependency: local_planner node owns a local_planner and a visualizer. After the local_planner has done a run, the local_planner_NODE passes the reference to the local_planner into the visualizer. I do not see a circle here

local_planner/src/nodes/local_planner_node.cpp
baumanta: because they are not local_planner specific. I wanted to have a node which does all the general stuf input/output to sensors and fcu and general SITL stuff. The visualizer is supposed to do all the local_planner specific visualization we need for debugging. Such that in far futur we could swap the local_planner with e.g. an ompl planner, reuse the same node with a different visualizer. But if those publishers annoy you where they are we can discuss putting the in the visualizer also, then we you just have to duplicate them for all visualizers (or have a base class visualizer, where all planners share this one)

local_planner/src/nodes/local_planner_visualization.cpp
```diff
@@ -0,0 +1,538 @@
...
+      nh.advertise<visualization_msgs::Marker>("/path_adapted_waypoint", 1);
+  current_waypoint_pub_ =
+      nh.advertise<visualization_msgs::Marker>("/current_setpoint", 1);
+  takeoff_pose_pub_ =
+      nh.advertise<visualization_msgs::Marker>("/take_off_pose", 1);
+  initial_height_pub_ =
+      nh.advertise<visualization_msgs::Marker>("/initial_height", 1);
+  histogram_image_pub_ =
+      nh.advertise<sensor_msgs::Image>("/histogram_image", 1);
+  cost_image_pub_ = nh.advertise<sensor_msgs::Image>("/cost_image", 1);
```
nicovanduijn: Did you go through each of these and make sure there are no legacy publishers that ceased to exist or be useful a while ago? This would be a great time to double check that!

PR #258, opened 2019-03-20
enable COM_OBS_AVOID by default

mrivi:
For the lazy people :)

### DIFF
--- a/global_planner/launch/mavros_sitl.launch
+++ b/global_planner/launch/mavros_sitl.launch
@@ -16,6 +16,7 @@
 
     <!-- Launch PX4 SITL -->
     <include file="$(find px4)/launch/px4.launch">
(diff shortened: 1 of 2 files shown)


nicovanduijn:
Wouldn't this require https://github.com/PX4/sitl_gazebo/pull/300 to be merged?

mrivi:
No, this is enough https://github.com/PX4/Firmware/blob/master/ROMFS/px4fmu_common/init.d-posix/1015_iris_obs_avoid

nicovanduijn:
So, there is no sdf model file for the iris_obs_avoid?

mrivi:
The vehicle being loaded is still the one in the avoidance models folder. I just want to load the iris default PX4 parameters plus `COM_OBS_AVOID` to 1 which is in set in that ROMFS script

PR #261, opened 2019-03-25
Don't calculate distance twice and init camera received

sunzj:
Since the distance of cloud points <--> position already calculate in cartesianToPolar(),don't calculate again.

Init camera_received to false in initializeCameraSubscribers()


baumanta:
I looked at it in SITL seems to run just fine

PR #264, opened 2019-03-26
Publish size of cropped point cloud

nicovanduijn:
This commit adds a new topic on which the planner publishes the
number of points in the cropped point cloud. This can be very
helpful when debugging on real hardware and logging capabilities are limited.


nicovanduijn:
Closing in favor of #255

PR #268, opened 2019-03-27
Trajectory simulation

jkflying:
This adds a simple trajectory simulator, which tries to mimic what the drone would do if given a direction setpoint. It only simulates position/velocity/acceleration, no orientation, uses jerk limiting and keeps within pre-defined thresholds which should be fairly compatible with the firmware parameters.

The idea is that it gives a list of all of the intermediate states, so that they can be tested for collisions with obstacles.

Not used anywhere yet, just the code and some tests added so far.

### DIFF
--- a/local_planner/include/local_planner/trajectory_simulator.h
+++ b/local_planner/include/local_planner/trajectory_simulator.h
@@ -19,10 +19,17 @@ struct simulation_limits {
   float max_jerk_norm = NAN;
 };
 
-std::vector<simulation_state> run_steps(const simulation_limits& config,
-                                        const simulation_state& start,
-                                        const Eigen::Vector3f& goal_direction,
-                                        float step_time, int num_steps);
+std::vector<simulation_state> velocity_trajectory(
+    const simulation_limits& config, const simulation_state& start,
+    const Eigen::Vector3f& goal_direction, float step_time, int num_steps);
+
+simulation_state simulate_step_constant_jerk(const simulation_state& state,
+                                             const Eigen::Vector3f& jerk,
+                                             float step_time);
+
+Eigen::Vector3f jerk_for_velocity_setpoint(
+    float P_constant, float D_constant, float max_jerk_norm,
+    const Eigen::Vector3f& desired_velocity, const simulation_state& state);
 
 template <int N>
 Eigen::Matrix<float, N, 1> norm_clamp(const Eigen::Matrix<float, N, 1>& val,
--- a/local_planner/include/local_planner/trajectory_simulator.h
+++ b/local_planner/include/local_planner/trajectory_simulator.h
@@ -1,0 +1,35 @@
+#pragma once
+
+#include <eigen3/Eigen/Core>
+
+namespace avoidance {
+
+struct simulation_state {
+  Eigen::Vector3f position = NAN * Eigen::Vector3f::Ones();
+  Eigen::Vector3f velocity = NAN * Eigen::Vector3f::Ones();
+  Eigen::Vector3f acceleration = NAN * Eigen::Vector3f::Ones();
+};
+
+struct simulation_limits {
+  float max_z_velocity = NAN;
+  float min_z_velocity = NAN;
+  float max_xy_velocity_norm = NAN;
+  float max_acceleration_norm = NAN;
+  float max_jerk_norm = NAN;
+};
+
+std::vector<simulation_state> run_steps(const simulation_limits& config,
+                                        const simulation_state& start,
+                                        const Eigen::Vector3f& goal_direction,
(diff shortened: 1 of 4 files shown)


local_planner/include/local_planner/trajectory_simulator.h
```diff
@@ -0,0 +1,36 @@
...
+
+std::vector<simulation_state> run_steps(const simulation_limits& config,
+                                        const simulation_state& start,
+                                        const Eigen::Vector3f& goal_direction,
+                                        float step_time, int num_steps);
+
+template <int N>
+Eigen::Matrix<float, N, 1> norm_clamp(const Eigen::Matrix<float, N, 1>& val,
+                                      float max_norm) {
+  float norm_sq = val.squaredNorm();
```
nicovanduijn: Is `squaredNorm()` defined for any N?

local_planner/include/local_planner/trajectory_simulator.h
```diff
@@ -0,0 +1,36 @@
...
+  if (norm_sq > max_norm * max_norm)
+    return val * (max_norm / std::sqrt(norm_sq));
+  else
+    return val;
+}
```
nicovanduijn: Could we move this to the cpp? Personally, I like everything that's more than just a getter or a setter to be defined in the cpp, but I guess this a matter of taste

Jaeyoung-Lim:
Pretty cool!

It would be interesting to see how close the vehicle behaves with the PD controller assumption done here

local_planner/include/local_planner/trajectory_simulator.h
```diff
@@ -0,0 +1,36 @@
+#pragma once
+
+#include <eigen3/Eigen/Core>
+
+namespace avoidance {
+
```
nicovanduijn: Just curious, what's the rationale behind the architecture here? Why individual functions as opposed to wrapping this in a class? Maybe it's just a brief draft and not thought through yet, maybe there's a reason I'm not aware of. I guess eventually we would want to include it in the planner through composition, as a 'Simulator' object or similar?

nicovanduijn:
Cool stuff! 
I'm curious to see how close this will be to the actual trajectory. We're essentially assuming saturation of the control limits (jerk, acceleration, velocity and eventually position), of which I'm not so sure if it's a good approximation. (Not that I have a better one in mind right now though)

local_planner/include/local_planner/trajectory_simulator.h
jkflying: I want to test it, so it needs something in the header. Also, templated functions have to be in the header so that each caller can get their own instantiation depending on what they call it with.

local_planner/include/local_planner/trajectory_simulator.h
jkflying: Good idea, I've arranged it to share the setup of the simulator so it can be initialized at each node.

local_planner/include/local_planner/trajectory_simulator.h
jkflying: Yep, internally it does `return numext::real((*this).cwiseAbs2().sum());`

jkflying:
Agreed with both @nicovanduijn  and @Jaeyoung-Lim about whether this follows the actual drone flight path. This is a good starting point, but I'm sure it will need to be tuned in the future. It is, however, a big improvement on not doing any vehicle dynamics modelling :wink:

Jaeyoung-Lim:
@jkflying I agree. Also the fact that multirotor systems are differentially flat and is usually limited to jerk limits, makes this a valid assumption.

local_planner/src/utils/trajectory_simulator.cpp
```diff
@@ -0,0 +1,107 @@
...
+                             unit_goal.z() > 0 ? config_.max_z_velocity
+                                               : config_.min_z_velocity),
+      config_.max_xy_velocity_norm, config_.min_z_velocity,
+      config_.max_z_velocity);
+
+  // calculate P and D constants such that they hit the jerk limit when
+  // doing accel from 0
+  float max_accel_norm = std::min(2 * std::sqrt(config_.max_jerk_norm),
+                                  config_.max_acceleration_norm);
+  float P_constant =
```
baumanta: This still bothers me a bit. It seems that P just needs to be something big and the relation between P and D has to be correct to be critically damped. But the actual value of P is not determined, this just sets a pseudo-scientific value for P. I wonder if there is an actual physically correct value, or if not if we should just put any high number instead.

jkflying: Agreed, let's look into this in the coming weeks.

PR #273, opened 2019-03-29
Enable velocity setpoints

jkflying:
This should be synchronized with this PR in the firmware: https://github.com/PX4/Firmware/pull/11561


jkflying:
Discussion at https://github.com/PX4/Firmware/pull/11561 , we will stay with position setpoints.

PR #274, opened 2019-03-29
Run Local Planner Node at a fixed rate

Jaeyoung-Lim:
This PR moves the while loop that lived in `local_planner_node_main` into `local_planner_node`, by using the `ros::AsyncSpinner`. This replaces #211 

- Using AsyncSpinner ensures the planner runs at a fixed specified rate
- This also enables `roscpp` to handle threading, which we can expand to other processes
- We can have less public members in the local planner node

I am not sure what to make of the `threadFunction`, should we also try to move this to a separate spinner?

This has been SITL tested.

### DIFF
--- a/local_planner/src/nodes/local_planner_node.cpp
+++ b/local_planner/src/nodes/local_planner_node.cpp
@@ -37,8 +37,6 @@ LocalPlannerNode::LocalPlannerNode(const ros::NodeHandle& nh,
 
   cmdloop_timer_ = nh_.createTimer(timer_options);
 
-  cmdloop_spinner_.start();
-
   // Set up Dynamic Reconfigure Server
   server_ = new dynamic_reconfigure::Server<avoidance::LocalPlannerNodeConfig>(
       config_mutex_, nh_);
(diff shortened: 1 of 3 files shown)


jkflying:
Looks good, let's get this flight tested.

Jaeyoung-Lim:
Rebased again due to a conflict introduced by #281

jkflying:
Flight test was good!

local_planner/src/nodes/local_planner_node.cpp
```diff
@@ -18,7 +18,7 @@ namespace avoidance {
 LocalPlannerNode::LocalPlannerNode(const ros::NodeHandle& nh,
                                    const ros::NodeHandle& nh_private,
                                    const bool tf_spin_thread)
-    : nh_(nh), nh_private_(nh_private) {
+    : nh_(nh), nh_private_(nh_private), spin_dt_(0.1) {
```
sunzj: 100ms? So is 10fps enough, the camera fps is 30fps.

Jaeyoung-Lim: It was 10fps before so first, it was not appropriate to change it in this PR. Second, replanning rate doesn't necessarily result in safer or smoother trajectories as the dynamics along the entire replanned trajectory is not that fast.

PR #278, opened 2019-04-01
use reprojection of combined histogram for starplanner instead of cropped cloud

baumanta:
old structure:

crop new cloud -> reproject points (old, combined histogram) -> build new hist and propagated hist -> combined hist -> star_planner (cropped cloud, reprojected points)

new structure:

crop new cloud -> build new hist and propagated hist -> combined hist -> reproject points (combined hist) -> star_planner (reprojected points)

This change allows the star planner to act on a tiny pointcloud (reprojected points of the combined histogram). The cropped cloud does not have to be passed into the starplanner anymore (one less copy) and the calculation per node get much simpler due to the smaller pointcloud it is using.

### DIFF
--- a/local_planner/src/nodes/local_planner.cpp
+++ b/local_planner/src/nodes/local_planner.cpp
@@ -110,7 +110,6 @@ void LocalPlanner::runPlanner() {
 void LocalPlanner::create2DObstacleRepresentation(const bool send_to_fcu) {
   // construct histogram if it is needed
   // or if it is required by the FCU
-  reprojectPoints(polar_histogram_);
   Histogram propagated_histogram = Histogram(2 * ALPHA_RES);
   Histogram new_histogram = Histogram(ALPHA_RES);
   to_fcu_histogram_.setZero();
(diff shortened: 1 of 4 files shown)


baumanta:
flight tests looked fine, no regression

local_planner/src/nodes/local_planner.cpp
```diff
@@ -126,6 +125,9 @@ void LocalPlanner::create2DObstacleRepresentation(const bool send_to_fcu) {
   }
   polar_histogram_ = new_histogram;
 
+  // create 3D points from combined histogram
+  reprojectPoints(polar_histogram_);
```
sunzj: That PR could save lots time in buildLookAheadTree().
However,since reprojectPoints histogram in the current stage,  the function reprojectPoints() should not use last position "position_old_", the reprojected histogram is based on current position.
https://github.com/PX4/avoidance/blob/4a92ae0a6fbfd3d654657808ecfa8c494149bcfa/local_planner/src/nodes/local_planner.cpp#L361-L364.

baumanta: good catch! I missed that one :)

PR #279, opened 2019-04-02
Transform pointclouds parallel

sunzj:
Transforming cloudpoint consume too much time,especially for multi-cameras,
so create threads to transform cloudpoints parallel.

### DIFF
--- a/local_planner/src/nodes/local_planner_node.cpp
+++ b/local_planner/src/nodes/local_planner_node.cpp
@@ -124,6 +124,12 @@ void LocalPlannerNode::initializeCameraSubscribers(
   std::vector<std::string> camera_info(camera_topics.size(), s);
 
   for (size_t i = 0; i < camera_topics.size(); i++) {
+    cameras_[i].trans_ready_mutex_.reset(new std::mutex);
+    cameras_[i].trans_ready_cv_.reset(new std::condition_variable);
+    cameras_[i].cloud_ready_mutex_.reset(new std::mutex);
+    cameras_[i].cloud_ready_cv_.reset(new std::condition_variable);
+    cameras_[i].transformed_ = false;
+
     cameras_[i].pointcloud_sub_ = nh_.subscribe<sensor_msgs::PointCloud2>(
         camera_topics[i], 1,
         boost::bind(&LocalPlannerNode::pointCloudCallback, this, _1, i));
@@ -142,6 +148,8 @@ void LocalPlannerNode::initializeCameraSubscribers(
     cameras_[i].camera_info_sub_ = nh_.subscribe<sensor_msgs::CameraInfo>(
         camera_info[i], 1,
         boost::bind(&LocalPlannerNode::cameraInfoCallback, this, _1, i));
+    cameras_[i].transform_thread_ =
+        std::thread(&LocalPlannerNode::pointCloudTransformThread, this, i);
   }
 }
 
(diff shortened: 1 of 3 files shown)


jkflying:
Hi @sunzj , have you done any benchmarking that you can share with us on how this improves speed? Transforming the cloud is pretty fast (although definitely a slow part of the algorithm), and I just want to make sure the overhead of context switching and running the multiple threads is worth the speed gain.

sunzj:
Hi @jkflying ,i don't know which platform you use, we run it on Qua-core 1.44G platform, it needs about 20ms to transform cloudpoints for one camera. 

The transform is done in updatePlanner(). If don't create threads to transform cloudpoints parallel, updatePlanner() consume about 20ms. With this PR, the function consume less than 1ms.

jkflying:
Thanks for the info! This really tells us where the bottleneck are in your setup.

What sizes are your point clouds? I really don't think the transforms should be so slow, we should investigate that and make it faster, having the processing in a different thread only moves the problem to a different core :slightly_smiling_face:

sunzj:
@jkflying i don't print the size of cloud points , but i test it on the same place, and don't move the drone.

It's not move the problem, it converts serial to parallel to speed up.
    Firstly, updatePlanner() will block path planner, thus cause path planner wait data ready, wast processing time.
    Secondly, when the cloudpoints came out from camera, no module handle/transform it. just wait,until updatePlanner() is called.

local_planner/src/nodes/local_planner_node.cpp
```diff
@@ -153,9 +161,21 @@ size_t LocalPlannerNode::numReceivedClouds() {
   return num_received_clouds;
 }
 
+size_t LocalPlannerNode::numTransformedClouds() {
+  size_t num_transformed_clouds = 0;
+  for (size_t i = 0; i < cameras_.size(); i++) {
+    std::unique_lock<std::mutex> lk(*(cameras_[i].trans_ready_mutex_));
+    cameras_[i].trans_ready_cv_->wait_for(
+        lk, std::chrono::milliseconds(30),
```
jkflying: Having this time hardcoded may cause problems on some systems. Is there some way we can do this without a time delay?

sunzj: For an asynchronous processing,it's common to use a waiting-time threshold to avoid hangup/block.

Since there are two threads,the main thread needs to wait cloudpoints transform thread finish its job. We can't assume it is called at the timing of "transforming just finish" every time. The worst case is: when wait_for is called, "cloudpoints transform thread" begin to transform cloudpoints. In such situation, the result is same as serial version. But if there are multi-cameras, it's still faster.

The hard code waiting-time threshold 30ms seems long. It won't wait such long time unless transform can't finish in 30ms which rarely happens.

jkflying:
Have you flight-tested this yet? 

Also, on what hardware exactly are you running? Something similar to Up Core?

local_planner/src/nodes/local_planner_node.cpp
jkflying: I think the delay is more likely to come from the transform not being available yet due to intermittent communication to the flight controller. But offloading something like that to another thread makes sense, so that the planner thread doesn't have `sleep`s in it.

sunzj:
Yes, i tested outdoor several times,it worked well.

we run it on Intel Qua-core 1.44G with Realsense R200

jkflying:
Great. We'll do some flight tests our side with D435i and the Up Core, also we need to test that it works well with multi-camera.

jkflying:
We tested it and it looked really good, we just need to add the fix for the box-size from yesterday and test again, then I'll be happy to merge this.

jkflying:
Flight testing with increased clipping radius worked really well. I've gone over the code in detail and everything looks correctly protected, each camera is in its own thread, etc, so this is good for merging.

Thanks for the contribution @sunzj!

jkflying:
Just for reference, here are the flight logs from the last test:

https://logs.px4.io/plot_app?log=82d915b0-3421-46f4-a872-584bb944c045
https://logs.px4.io/plot_app?log=a86f3040-ed89-40d5-94a6-305a5748f039

sunzj:
@jkflying Thanks for your test! and I also test with increased box size, the drone react far away from obstacle, it will improve drone's speed.

PR #281, opened 2019-04-02
Use firmware parameters

mrivi:
- rename parameters using Firmware convention
- use MPC_XY_CRUISE as max velocity when there aren't any obstacles
- use MPC_COL_PREV_D to decide if the distance data are sent to the Firmware
- remove unused dynamic reconfigure paramters
- add MPC_JERK_MAX since only the min was retrieved

### DIFF
--- a/local_planner/cfg/local_planner_node.cfg
+++ b/local_planner/cfg/local_planner_node.cfg
@@ -34,13 +34,4 @@ gen.add("tree_node_distance_",    double_t,    0, "Distance between nodes", 1,  
 gen.add("tree_discount_factor_",    double_t,    0, "Discount factor in tree cost function", 0.8,  0, 1)
 gen.add("max_path_length_",    double_t,    0, "Maximum length of planned paths", 3,  0, 15)
 
-# waypoint_generator
-gen.add("goal_acceptance_radius_in_", double_t, 0, "Radius of a sphere around a waypoint to set it as reached", 0.5,  0.1, 5.0)
-gen.add("goal_acceptance_radius_out_", double_t, 0, "Radius of a sphere around a waypoint to set it as not reached", 1.5,  0.1, 10.0)
-gen.add("factor_close_to_goal_start_speed_limitation_", double_t, 0, "", 3.0,  0.1, 20.0)
-gen.add("factor_close_to_goal_stop_speed_limitation_", double_t, 0, "", 4.0,  0.1, 20.0)
-gen.add("max_speed_close_to_goal_factor_", double_t, 0, "", 0.1,  0.1, 20.0)
-gen.add("min_speed_close_to_goal_", double_t, 0, "Speed close to goal", 0.5,  0.1, 14.0)
-
-
 exit(gen.generate(PACKAGE, "avoidance", "LocalPlannerNode"))
--- a/local_planner/cfg/local_planner_node.cfg
+++ b/local_planner/cfg/local_planner_node.cfg
@@ -12,7 +12,7 @@ gen.add("goal_cost_param_", double_t, 0, "Cost function weight for goal oriented
 gen.add("heading_cost_param_", double_t, 0, "Cost function weight for constant heading", 0.5, 0, 20.0)
 gen.add("smooth_cost_param_", double_t, 0, "Cost function weight for path smoothness", 1.5, 0.5, 20.0)
 gen.add("velocity_around_obstacles_", double_t, 0, "Maximally allowed drone velocity in the proximity of obstacles [m/s]", 2.0, 0.0, 5.0)
-gen.add("velocity_far_from_obstacles_", double_t, 0, "Maximally allowed drone velocity when no obstacles are in sight [m/s]", 3.0, 0.0, 10.0)
+gen.add("keep_distance_", double_t, 0, "Distance at which the UAV stops in front of an obstacle", 2.0, 1.0, 8.0)
 gen.add("goal_z_param", double_t, 0, "Height of the goal position", 3.5, -20.0, 20.0)
 gen.add("no_progress_slope_", double_t, 0, "If progress derivative higher than this value the drone rises", -0.0007, -1.0, 1.0)
 gen.add("min_realsense_dist_", double_t, 0, "Discard points closer than that", 0.2, 0, 10)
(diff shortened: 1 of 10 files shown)


local_planner/cfg/local_planner_node.cfg
```diff
@@ -32,7 +31,6 @@ gen.add("stop_in_front_", bool_t, 0, "Enable stop in front of the obstacle", Fal
 gen.add("use_back_off_", bool_t, 0, "Enable functionality to move backwards if an obstacle is too close", False)
 gen.add("use_VFH_star_", bool_t, 0, "Build lookahead-tree", True)
```
nicovanduijn: I've been meaning to get rid of these two as well. We pretty much always use VFH-star these days, and if someone wants to really only plan one node, they could set `n_expanded_nodes` to 1, which should have the same effect as setting `use_VFH_star` to `False`.
The back-off has also been disabled by default for a while, do you have an opinion on that? Can we get rid of it?

local_planner/cfg/local_planner_node.cfg
```diff
@@ -12,7 +12,6 @@ gen.add("goal_cost_param_", double_t, 0, "Cost function weight for goal oriented
 gen.add("heading_cost_param_", double_t, 0, "Cost function weight for constant heading", 0.5, 0.5, 20.0)
 gen.add("smooth_cost_param_", double_t, 0, "Cost function wight for path smoothness", 1.5, 0.5, 20.0)
 gen.add("velocity_around_obstacles_", double_t, 0, "Maximally allowed drone velocity in the proximity of obstacles [m/s]", 2.0, 0.0, 5.0)
-gen.add("velocity_far_from_obstacles_", double_t, 0, "Maximally allowed drone velocity when no obstacles are in sight [m/s]", 3.0, 0.0, 10.0)
 gen.add("keep_distance_", double_t, 0, "Distance at which the UAV stops in front of an obstacle", 2.0, 1.0, 8.0)
```
nicovanduijn: My first thought here was that it could be replaced with `MPC_COL_PREV_D`. But I think this is related to the `stop_in_front` functionality? Another legacy piece of code I've been hoping to get rid of  (would then trim down the messy `determineStrategy()`, and also remove the `stop_in_front_` in line 31)

local_planner/src/nodes/local_planner.cpp
```diff
@@ -65,7 +63,7 @@ void LocalPlanner::dynamicReconfigureSetParams(
   use_back_off_ = config.use_back_off_;
   use_VFH_star_ = config.use_VFH_star_;
   adapt_cost_params_ = config.adapt_cost_params_;
-  send_obstacles_fcu_ = config.send_obstacles_fcu_;
+  send_obstacles_fcu_ = model_params_.param_mpc_col_prev_d > 0.f;
```
nicovanduijn: that's a bit wordy now, no? Is this the convention in the Firmware? Personally, I think the `model_params_` encodes the fact that they're params, so the members themselves don't really need to repeat that

local_planner/cfg/local_planner_node.cfg
```diff
@@ -41,13 +39,4 @@ gen.add("tree_node_distance_",    double_t,    0, "Distance between nodes", 1,
...
 
-# waypoint_generator
-gen.add("goal_acceptance_radius_in_", double_t, 0, "Radius of a sphere around a waypoint to set it as reached", 0.5,  0.1, 5.0)
-gen.add("goal_acceptance_radius_out_", double_t, 0, "Radius of a sphere around a waypoint to set it as not reached", 1.5,  0.1, 10.0)
-gen.add("factor_close_to_goal_start_speed_limitation_", double_t, 0, "", 3.0,  0.1, 20.0)
-gen.add("factor_close_to_goal_stop_speed_limitation_", double_t, 0, "", 4.0,  0.1, 20.0)
-gen.add("max_speed_close_to_goal_factor_", double_t, 0, "", 0.1,  0.1, 20.0)
-gen.add("min_speed_close_to_goal_", double_t, 0, "Speed close to goal", 0.5,  0.1, 14.0)
-
-
```
nicovanduijn: nice!

local_planner/src/nodes/local_planner.cpp
mrivi: I really like the idea of keeping the same naming convention of the Firmware especially since it's going to be strictly enforced (https://github.com/PX4/Firmware/pull/11686).
What if we change `model_params_` to `px4`?

local_planner/cfg/local_planner_node.cfg
mrivi: yes, I think so. @baumanta @jkflying your opinions?

local_planner/src/nodes/local_planner.cpp
nicovanduijn: Sure, even better! `model_params` isn't very clear to me anyway

local_planner/cfg/local_planner_node.cfg
jkflying: Agreed.

jkflying:
It is more important that these are correct now that we will be using them. Is there any way we can detect if the parameters didn't arrive? Maybe the default values should be NaN, and after some delay if any are still NaN we should ask for the values again.

Otherwise I'm happy with this, just needs a formatting fix.

local_planner/cfg/local_planner_node.cfg
mrivi: let's do it in a separate PR

local_planner/include/local_planner/local_planner.h
```diff
@@ -46,23 +46,23 @@ class TreeNode;
...
-  float param_mpc_tko_speed = 1.0f; // Takeoff climb rate
-  float param_mpc_land_speed = 0.7f;   // Landing descend rate
+  float param_mpc_xy_cruise = NAN;   // Desired horizontal velocity in mission
+  float param_mpc_tko_speed = NAN; // Takeoff climb rate
+  float param_mpc_land_speed = NAN;   // Landing descend rate
   // limitations given by sensors
-  float param_ekf2_rng_a_hmax = 5.0f;
-  float param_ekf2_rng_a_vmax = 5.0f;
+  float param_ekf2_rng_a_hmax = NAN;
+  float param_ekf2_rng_a_vmax = NAN;
```
mhkabir: EKF2 Range aid parameters are not the correct ones to use to get sensor limitations.

Ideally we would transport and use these limits computed by the EKF: https://github.com/PX4/Firmware/blob/master/msg/vehicle_local_position.msg#L60

jkflying:
I think the retry if we don't get the parameters is with this: http://docs.ros.org/api/mavros_msgs/html/srv/ParamGet.html

We can get just the parameters we need if they don't arrive after a timeout. It is a service, so it won't have any dropped packets, too.

jkflying:
Flight testing for this went really well. Once thing we found is that most of the time it spends at the 1/2 max-velocity due to some obstacle appearing somewhere in the history. As a result the actual max speed isn't exercised very often. Due to this, to make the flight more predictable it would be a good idea to get rid of the 1-tier speeds, rather just using a single speed and depending on the flight planning to slow down the drone when rapid trajectory changes are needed.

PR #282, opened 2019-04-03
use the current position in reprojection

baumanta:
Bug was introduced in [#278](https://github.com/PX4/avoidance/pull/278) found by @sunzj 

As now the current histogram is reprojected, also the current position should be used.


nicovanduijn:
A more descriptive title would be nice, such that the git history is somewhat meaningful.
Otherwise I'm fine with this. Can you notice the difference in SITL?

baumanta:
@nicovanduijn  no, I don't see a difference, the effect is minimal.

jkflying:
I think this is small enough, SITL testing should be sufficient.

PR #284, opened 2019-04-04
Remove legacy code

mrivi:
Removes:
- goBack
- stopInFront
- separate case for not using VFH*

### DIFF
--- a/local_planner/src/nodes/local_planner.cpp
+++ b/local_planner/src/nodes/local_planner.cpp
@@ -164,23 +164,21 @@ void LocalPlanner::determineStrategy() {
       create2DObstacleRepresentation(true);
     }
   } else {
-    evaluateProgressRate();
+    waypoint_type_ = tryPath;
 
+    evaluateProgressRate();
     create2DObstacleRepresentation(send_obstacles_fcu_);
 
     // decide how to proceed
     if (hist_is_empty_) {
       obstacle_ = false;
-      waypoint_type_ = tryPath;
-    }
-
-    if (!hist_is_empty_ && reach_altitude_) {
+    } else {
       obstacle_ = true;
 
-      getCostMatrix(
-          polar_histogram_, goal_, position_, curr_yaw_histogram_frame_deg_,
-          last_sent_waypoint_, cost_params_, velocity_.norm() < 0.1f,
-          smoothing_margin_degrees_, cost_matrix_, cost_image_data_);
+      getCostMatrix(polar_histogram_, goal_, position_,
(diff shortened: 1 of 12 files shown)


local_planner/src/nodes/planner_functions.cpp
```diff
@@ -12,16 +12,12 @@ namespace avoidance {
 // considered
 void filterPointCloud(
```
jkflying: @baumanta heads up, this might conflict with what you are working on

local_planner/src/nodes/local_planner.cpp
```diff
@@ -169,97 +160,42 @@ void LocalPlanner::determineStrategy() {
...
-             reach_altitude_) {
-    obstacle_ = true;
-    ROS_INFO(
-        "\033[1;35m[OA] There is an Obstacle Ahead stop in front\n \033[0m");
-    stopInFrontObstacles();
-    waypoint_type_ = direct;
-
     if (send_obstacles_fcu_) {
       create2DObstacleRepresentation(true);
     }
```
nicovanduijn: I'm guessing that 
```
    if (send_obstacles_fcu_) {
      create2DObstacleRepresentation(true);
    }
```
is the same as `create2DObstacleRepresentation(send_obstacles_fcu_)` ?

It seems that `create2DObstacleRepresentation(send_obstacles_fcu_)` is done both in the `if(!reach_altitude_)` and it's `else...`. Can we just do it once in the beginning?

local_planner/src/nodes/local_planner.cpp
```diff
@@ -169,97 +160,42 @@ void LocalPlanner::determineStrategy() {
...
+
+    evaluateProgressRate();
+    create2DObstacleRepresentation(send_obstacles_fcu_);
 
+    // decide how to proceed
+    if (hist_is_empty_) {
+      obstacle_ = false;
     } else {
-      evaluateProgressRate();
+      obstacle_ = true;
```
nicovanduijn: It looks like `obstacle_` and `hist_is_emtpy_` essentially contain the same information. Can we get rid of one of them?

local_planner/src/nodes/local_planner.cpp
mrivi: You would compute the histograms for no reasons during takeoff if you don't need to send the data to the Firmware

mrivi:
yes, I didn't notice any difference with respect to master

baumanta:
successful flight tests, no regression

PR #285, opened 2019-04-04
Fix auto takeoff

mrivi:
Auto takeoff from QGC was not working when the OA was running.
This is because the navigation state is AUTO_TAKEOFF and not MISSION.

### DIFF
--- a/local_planner/include/local_planner/local_planner_node.h
+++ b/local_planner/include/local_planner/local_planner_node.h
@@ -113,6 +113,7 @@ enum class NavigationState {
   mission,
   auto_takeoff,
   auto_land,
+  auto_rtl,
   auto_rtgs,
   offboard,
   none,
--- a/local_planner/include/local_planner/local_planner_node.h
+++ b/local_planner/include/local_planner/local_planner_node.h
@@ -113,6 +113,7 @@ enum class NavigationState {
   mission,
   auto_takeoff,
   auto_land,
+  auto_rtgs,
   offboard,
   none,
 };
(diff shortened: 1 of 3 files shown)


local_planner/include/local_planner/local_planner_node.h
```diff
@@ -275,6 +275,7 @@ class LocalPlannerNode {
 
   geometry_msgs::TwistStamped vel_msg_;
   bool armed_, offboard_, mission_, new_goal_;
+  bool auto_takeoff_;
```
jkflying: There's no way these uninitialized variables can be read before they are set, right?

Maybe we should set them all to `false` by default?

local_planner/src/nodes/local_planner_node.cpp
```diff
@@ -280,17 +280,25 @@ void LocalPlannerNode::stateCallback(const mavros_msgs::State& msg) {
...
     offboard_ = false;
     mission_ = false;
+    auto_takeoff_ = false;
   }
 }
 
 void LocalPlannerNode::calculateWaypoints(bool hover) {
-  bool is_airborne = armed_ && (mission_ || offboard_ || hover);
+  bool is_airborne =
+      armed_ && (mission_ || offboard_ || auto_takeoff_ || hover);
```
jkflying: Is there maybe a similar case for landing or getting the drone the disarm?

local_planner/include/local_planner/local_planner_node.h
```diff
@@ -109,6 +109,16 @@ enum class MAV_STATE {
...
+enum class NavigationState {
+  mission,
+  auto_takeoff,
+  auto_land,
+  auto_rtl,
+  auto_rtgs,
+  offboard,
+  none,
+};
+
```
Jaeyoung-Lim: Nice!

Jaeyoung-Lim: Nice!

PR #286, opened 2019-04-04
Change recursion strategy

baumanta:
Change the way old data is remembered. The old recursion strategy saved the last histogram and then transformed it back into a pointcloud. This way a lot of discretization error was accumulated due to going from histogram to pointcloud and back in every computation cycle. This rendered the remembered date invalid after about 30 computation cycles.

The proposed solution saves a subsampled version of the original data points. The pointcloud from the last timestep is then fuse with the new data. This yields much better results.

With this method the planner successfully completes the test mission in all runs, with one camera (depth or realsense) as well as with 3 cameras. 

recursion_strategy_comparison


local_planner/src/nodes/planner_functions.cpp
```diff
@@ -10,19 +10,26 @@ namespace avoidance {
...
-    int min_cloud_size, float min_dist_backoff, Box histogram_box,
-    const Eigen::Vector3f& position, float min_realsense_dist) {
-  cropped_cloud.points.clear();
-  cropped_cloud.width = 0;
+    float min_dist_backoff, Box histogram_box, const Eigen::Vector3f& position,
+    float min_realsense_dist, int max_age) {
+  pcl::PointCloud<pcl::PointXYZI> old_cloud = final_cloud;
+
+  final_cloud.points.clear();
+  final_cloud.width = 0;
```
nicovanduijn: could you reserve the size to the maximum it could be? I think that would be faster than dynamic reallocation

local_planner/src/nodes/planner_functions.cpp
```diff
@@ -45,13 +59,36 @@ void filterPointCloud(
...
+    // cloud
+    if (histogram_box.isPointWithinBox(xyzi.x, xyzi.y, xyzi.z)) {
+      distance = (position - toEigen(xyzi)).norm();
+      if (distance < histogram_box.radius_) {
+        PolarPoint p_pol = cartesianToPolar(toEigen(xyzi), position);
+        Eigen::Vector2i p_ind = polarToHistogramIndex(p_pol, ALPHA_RES / 2);
+        if (high_res_histogram.get_age(p_ind.y(), p_ind.x()) == 0 &&
+            xyzi.intensity < max_age) {
+          final_cloud.points.push_back(
+              toXYZI(xyzi.x, xyzi.y, xyzi.z, xyzi.intensity + 1));
```
nicovanduijn: As discussed, the intensity layer is in float, so this increment is problematic. Switching to using floats and actual age in seconds

local_planner/src/nodes/planner_functions.cpp
```diff
@@ -45,13 +59,36 @@ void filterPointCloud(
...
+  // combine with old cloud
+  for (const pcl::PointXYZI& xyzi : old_cloud) {
+    // adding older points if not expired and space is free according to new
+    // cloud
+    if (histogram_box.isPointWithinBox(xyzi.x, xyzi.y, xyzi.z)) {
+      distance = (position - toEigen(xyzi)).norm();
+      if (distance < histogram_box.radius_) {
+        PolarPoint p_pol = cartesianToPolar(toEigen(xyzi), position);
+        Eigen::Vector2i p_ind = polarToHistogramIndex(p_pol, ALPHA_RES / 2);
+        if (high_res_histogram.get_age(p_ind.y(), p_ind.x()) == 0 &&
```
nicovanduijn: You are now using the "age" layer of the histogram for something completely new - it's not containing the age but the binary information whether we have a point in this cell yet. This is fine, but you should rename it in a meaningful way in order to keep things maintainable and understandable down the road

local_planner/src/nodes/planner_functions.cpp
jkflying: Agreed, `final_cloud.points.reserve(histogram.rows() * histogram.cols());`

Repeatedly reserving (and then destroying) the exact same sized chunk of memory also makes things easier for the memory allocator to re-use memory.

local_planner/src/nodes/planner_functions.cpp
```diff
@@ -10,19 +10,26 @@ namespace avoidance {
...
     Eigen::Vector3f& closest_point, float& distance_to_closest_point,
     int& counter_backoff,
     const std::vector<pcl::PointCloud<pcl::PointXYZ>>& complete_cloud,
-    int min_cloud_size, float min_dist_backoff, Box histogram_box,
-    const Eigen::Vector3f& position, float min_realsense_dist) {
-  cropped_cloud.points.clear();
-  cropped_cloud.width = 0;
+    float min_dist_backoff, Box histogram_box, const Eigen::Vector3f& position,
+    float min_realsense_dist, int max_age) {
+  pcl::PointCloud<pcl::PointXYZI> old_cloud = final_cloud;
```
jkflying: Rather than copying and then resetting, simply do:
```
pcl::PointCloud<pcl::PointXYZI> old_cloud;
std::swap(final_cloud, old_cloud);
```

This avoids the copy of the points.

local_planner/src/nodes/planner_functions.cpp
```diff
@@ -31,7 +38,14 @@ void filterPointCloud(
           distance = (position - toEigen(xyz)).norm();
           if (distance > min_realsense_dist &&
               distance < histogram_box.radius_) {
-            cropped_cloud.points.push_back(pcl::PointXYZ(xyz.x, xyz.y, xyz.z));
+            // subsampling the cloud
+            PolarPoint p_pol = cartesianToPolar(toEigen(xyz), position);
+            Eigen::Vector2i p_ind = polarToHistogramIndex(p_pol, ALPHA_RES / 2);
+            if (high_res_histogram.get_age(p_ind.y(), p_ind.x()) == 0) {
+              final_cloud.points.push_back(toXYZI(xyz.x, xyz.y, xyz.z, 0));
```
jkflying: Maybe easier to do 
`final_cloud.points.push_back(toXYZI(toEigen(xyz), 0));`

local_planner/src/nodes/planner_functions.cpp
```diff
@@ -31,7 +38,14 @@ void filterPointCloud(
...
+              high_res_histogram.set_age(p_ind.y(), p_ind.x(), 1);
```
jkflying: I'm just thinking, now that we store the age in the points, do we still need the age in the histogram? What purpose does it serve anymore?

local_planner/src/nodes/planner_functions.cpp
jkflying: Maybe better would be to keep NaN in the distance matrix if there is no sample, rather than a separate layer.

local_planner/src/nodes/planner_functions.cpp
jkflying: What he said

local_planner/src/nodes/planner_functions.cpp
baumanta: removed age layer from histogram class, as age is now stored in the pointcloud per point. Stuck to using 0 for empty cells instead of NANs, as we add distances to bins and divide by n points. This would get much more complicated it empty cells were NAN

jkflying:
Flight test looked really good.

PR #289, opened 2019-04-05
add min number of points per cell to reduce noise

baumanta:
this is a new version of [PR 223](https://github.com/PX4/avoidance/pull/223)

It should mitigate noise in the incoming data by only storing point, which have enough other ones confirming them.

The exact parameter value has to be tune in test flights


jkflying:
Flight testing of this showed problems, we'll have to come back to this at some point but for now this can be closed.

PR #293, opened 2019-04-08
Set final cloud info finally

sunzj:
Set the final_cloud info outside the loop.
Early stopping while check histogram empty.


local_planner/include/local_planner/histogram.h
```diff
@@ -55,7 +56,10 @@ class Histogram {
   * @param[in] value, distance to the vehicle of obstacle mapped to (x, y) cell
   *[m]
   **/
-  inline void set_dist(int x, int y, float value) { dist_(x, y) = value; }
+  inline void set_dist(int x, int y, float value) {
+    dist_(x, y) = value;
+    is_empty = false;
```
baumanta: What if a cell is set back to zero? 
We fill all histogram cells with the sum of all distances, then , if not enough points are in one cell it is set back to 0 as we assume it was noise. I think this will not work for this case: https://github.com/PX4/avoidance/blob/978bb05f3fd6b060de4c417809e73318a3e121a4/local_planner/src/nodes/planner_functions.cpp#L134

sunzj: although now don't use the "noise" condition https://github.com/PX4/avoidance/blob/978bb05f3fd6b060de4c417809e73318a3e121a4/local_planner/src/nodes/planner_functions.cpp#L130
In case sometime use it. ...So still need check all histogram cells. I rollback to early stopping.

local_planner/src/nodes/histogram.cpp
```diff
@@ -52,17 +53,10 @@ void Histogram::downsample() {
...
-      }
-    }
-  }
-  return counter == 0;
+void Histogram::setZero() {
+  dist_.fill(0.f);
+  is_empty_ = true;
 }
+
+bool Histogram::isEmpty() const { return is_empty_; }
```
Jaeyoung-Lim: Correct me if I am wrong, but I think this will never become false with the current implementation.

baumanta: yes that was exactly was I was trying to say :+1:

sunzj: Since the cell may set to 0.f. So it needs check all cells in Histogram::isEmpty(), rollback to early stopping version.

baumanta:
@jkflying @sunzj I think that would be a good idea. There are already histogram test in here https://github.com/sunzj/avoidance/blob/90f96eaa11e71ab75620c683f7d2d08719ca4dcc/local_planner/test/test_planner_functions.cpp#L701 I think you could just add the `isEmpty` test there as well

sunzj:
@jkflying @baumanta added unit test for Histogram::isEmpty().

PR #304, opened 2019-04-15
Generate histogram when process cloud points

sunzj:
Generate histogram when process clouds,  and pass it to starplanner.


jkflying:
Hi, what is the reason for making the polar histogram outside?

sunzj:
@jkflying To make it more efficient.

Since polar histogram/final cloud points are made from the raw cloud points, they can be generated at the same place in processPointcloud, then don't need to read these cloud points again.

And for buildLookAheadTree in star planner, the first polar histogram is the same one generated in local planner, so don't re-create it.

jkflying:
My issue with this is that we now have 2 separate places where we generate the histogram, so any changes we do we will need to remember to do it in 2 places, and to keep the exactly the same. Also, the pointcloud is very small now so this is histogram generation is now very fast, so if we do it twice for the first node it isn't really a problem.

sunzj:
@jkflying emm... it's not a problem, result is same. But it's unnecessary,we don't need read the cloud points twice or re-create the same histogram.

jkflying:
I'd rather not mix in two different algorithms into the same loop, it is ok while that is what we are working on but will be difficult in the future for bugfixes and new features.

Also I don't think that the performance is a big issue, since the point cloud is so small now.

PR #305, opened 2019-04-15
Offtrack State

mrivi:
This PR is an effort to make the avoidance system seamless to the mission mode. When the vehicle has finished avoiding an obstacle and it's more than the cruise speed away from the line previous-current goal, it will go back to the line between the previous and the current gaol entering at 60 degrees. 

offtrack


jkflying:
Needs a rebase

mrivi:
Rebased

jkflying:
Re-tested this in SITL, I think as an initial implementation it will be ok, we might need further ideas for how to do this during the avoidance mode as well.

PR #306, opened 2019-04-15
Verbosity

baumanta:
Add debugging info if the planner fails to obtain position data from the FCU
debugging

### DIFF
--- a/local_planner/resource/custom_rosconsole.conf
+++ b/local_planner/resource/custom_rosconsole.conf
@@ -1,1 +1,1 @@
-log4j.logger.ros.local_planner=ERROR
+log4j.logger.ros.local_planner=WARN
--- a/local_planner/src/nodes/local_planner_node_main.cpp
+++ b/local_planner/src/nodes/local_planner_node_main.cpp
@@ -46,24 +46,27 @@ int main(int argc, char** argv) {
 
     ros::Duration timeout_termination =
           ros::Duration(Node.local_planner_->timeout_termination_);
+    ros::Time start_query_position = ros::Time::now();
+    bool sent_error = false;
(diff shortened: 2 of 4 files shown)


local_planner/src/nodes/local_planner_node_main.cpp
```diff
@@ -42,9 +42,30 @@ int main(int argc, char** argv) {
...
     while (!Node.position_received_ && ros::ok()) {
       ros::getGlobalCallbackQueue()->callAvailable(ros::WallDuration(0.1));
+      ros::Duration since_query = ros::Time::now() - start_query_position;
+      if (since_query > timeout_termination && !sent_error) {
+        // clang-format off
+    	  ROS_WARN("\033[1;33m Planner abort: missing required data \n \033[0m");
+    	  ROS_WARN("----------------------------- Debugging Info -----------------------------");
+    	  ROS_WARN("Local planner has not received a position from FCU, check the following: ");
+    	  ROS_WARN("1. Check cables connecting PX4 autopilot with onboard computer");
+    	  ROS_WARN("2. Set PX4 parameter MAV_1_MODE to Onbard");
```
jkflying: *onboard

local_planner/resource/custom_rosconsole.conf
```diff
@@ -1 +1 @@
-log4j.logger.ros.local_planner=ERROR
\ No newline at end of file
+log4j.logger.ros.local_planner=WARN
```
jkflying: We aren't using WARN anywhere else that might get noisy now, are we?

baumanta: now, it is just this paragraph and the missing pointcloud that are WARN

PR #307, opened 2019-04-15
Fix ObstacleDistance min and max range 

mrivi:
Set obstacle distance ranges to what we actually use in the histogram


baumanta:
we flight tested this and it worked fine:
https://logs.px4.io/plot_app?log=6396b1c2-7985-4030-861a-85fead6c111f

PR #308, opened 2019-04-16
Changes for compatibility with new librealsense

jkflying:
Remove parameters which were removed in the Realsense ROS wrapper, which is required for the new librealsense.

Tested against librealsense 2.20.0 and the intel-ros/realsense 2.2.3


jkflying:
Also tested against older librealsense and intel-ros/realsense, no regressions.

PR #310, opened 2019-04-17
Fov empty priority

baumanta:
This PR should reduce noise by giving the new data absolute priority inside the FOV. This means that inside the FOV the remembered points are not taken into account. If there is an empty spot inside the FOV it will stay empty even if the remembered points say otherwise.


jkflying:
Flight test went well, @sunzj this might fix your outdoor test flight issues.

PR #312, opened 2019-04-17
Stop storing status_msg as status

Jaeyoung-Lim:
This PR stops the system status being stored as a companionstatus message, but directly on MAV_STATE

- This makes changing the system state more simpler as you don't need to understand the companionStatus message structure
- This fits better with the implemenation in global planner https://github.com/PX4/avoidance/blob/e723f87c18723954c9a8f8ec8974581b6cc89bcb/global_planner/src/nodes/path_handler_node.cpp#L269-L275

### DIFF
--- a/local_planner/include/local_planner/local_planner_node.h
+++ b/local_planner/include/local_planner/local_planner_node.h
@@ -96,8 +96,6 @@ class LocalPlannerNode {
                    const bool tf_spin_thread = true);
   ~LocalPlannerNode();
 
-  mavros_msgs::CompanionProcessStatus status_msg_;
-
   std::string world_path_;
   std::atomic<bool> should_exit_{false};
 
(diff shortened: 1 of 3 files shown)


local_planner/src/nodes/local_planner_node_main.cpp
```diff
@@ -14,27 +14,7 @@ int main(int argc, char** argv) {
   std::thread worker(&LocalPlannerNode::threadFunction, &Node);
 
   std::thread worker_params(&LocalPlannerNode::checkPx4Parameters, &Node);
-  ros::Duration timeout_termination =
```
jkflying: These helper warnings need to be added back

Jaeyoung-Lim: The helper warnings here are a artifact from a merge introduced by #274 The helper warnings are still in `local_planner_node.cpp`

Jaeyoung-Lim: Rebased just to make it clear

Jaeyoung-Lim:
Tested rebase in SITL

PR #313, opened 2019-04-17
Get rid of unnecessary sleep

Jaeyoung-Lim:
This PR gets rid of a 2 second sleep between starting the node and the threads.

I don't think this sleep is necessary anymore, since we don't need to wait for the node to come up.

I Have tested this in SITL. As far as I can see, this will only delay the ROS node from starting.


jkflying:
Looks good!

PR #315, opened 2019-04-18
Base the position of closed node to create histogram

sunzj:
When expand nodes, should base the position of closed node, seem there is a typo.
@baumanta


baumanta:
@sunzj you are right! thanks for looking so closely :)

sunzj:
@baumanta You are welcome!

PR #317, opened 2019-04-19
Early stopping if can't find expanded node

sunzj:
*Define variable outside the loop.
*Early stopping if can't find expanded node.
*Change the mode of source file to 644 in fix_style.

Gazebo Simulation test pass.

### DIFF
--- a/local_planner/src/nodes/star_planner.cpp
+++ b/local_planner/src/nodes/star_planner.cpp
@@ -88,6 +88,7 @@ float StarPlanner::treeCostFunction(int node_number) const {
                   static_cast<float>(tree_[node_number].depth_)) *
          (target_cost + smooth_cost + smooth_cost_to_old_tree + turning_cost);
 }
+
 float StarPlanner::treeHeuristicFunction(int node_number) const {
   Eigen::Vector3f node_position = tree_[node_number].getPosition();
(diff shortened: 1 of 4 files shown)


local_planner/src/nodes/star_planner.cpp
```diff
@@ -109,6 +110,27 @@ float StarPlanner::treeHeuristicFunction(int node_number) const {
 
 void StarPlanner::buildLookAheadTree() {
   std::clock_t start_time = std::clock();
+
+  int origin = 0, old_origin = 0;
```
jkflying: Hello

Having all of the variables declared before they are used doesn't actually improve performance, especially for things like `int`, `float` and other types that don't use heap allocation. Modern compilers will correctly allocate the space for the variable outside of the loop if it will cause a speedup, or not if it will slow down the code.

Especially with C++, it is important to keep the code more readable, and let the compiler do the optimization work.

Could you then put the variable declarations back to where they are used? The only places I see it making sense to declare at the front is something like `std::vector` or `Eigen::Matrix`, if it will be resized multiple times during the lifetime of the variable.

sunzj: Done

jkflying:
Checked after rebase in SITL with a mission, all works well.

Issue #309, opened 2019-04-16
Out door test fail after integrate "Change recursion strategy #286" with Realsense R200


sunzj:
@baumanta 

After integrate https://github.com/PX4/avoidance/pull/286, out door test fail. I have tested several times,the drone didn't follow the "mission path" even there wasn't any obstacle.

However, if use the previous  PR https://github.com/PX4/avoidance/pull/284, it works well.

So the result is, out door test. PR 284 work well, PR 284 + PR 286 fail.

Since i use realsense R200, the cloudpoints may not good enough. Did you verify on realsense R200 or other cameras?

And i find the final cloud store more than 20 seconds cloud points, it may have too much noise in it for R200.

jkflying:
I think the problem here is the duration of the history, combined with the big clipping radius, gives lots of noise. Have you tried reducing these two? It should be easy to change the defaults in the rqt_reconfigure.

We are testing with the D435, I'm not sure of the noise/performance compared to the R200, but maybe you can load the 'accurate' profile and it will have less noise. This is what we did for the D435.

jkflying:
There is also the possibility that previously any noise in the camera FOV would be deleted and replaced with the new camera data, since it was a histogram. This no longer happens, since it is a point cloud. We could definitely add this back.

sunzj:
i don't have chance to do outdoor test now, i'll share result once test outdoor with reducing "duration of the history" and "clipping radius"

sunzj:
Test outdoor, set the max_age to 20 frames, box_radius_ to 6,the issue still exist.

jkflying:
Hi, did PR310 fix the issue?

sunzj:
@jkflying no,PR310 doesn't work, but the result was a little better, the drone followed the "mission path" a short time, then stopped, and actually there wasn't any obstacle.

and i also tried PR 286 + PR310 + noise-supression-rebased, set max_age to 8, min_num_points_per_cell_ to 12, box_radius_ to 6, the result was much better,it stopped at the front of obstacle a long time, yawed left/right , finally found a path.

nicovanduijn:
Do you have a flight log or a rosbag or something? I'm curious as to what may be the root cause

sunzj:
@jkflying @nicovanduijn  finally it works, decrease goal_cost_param_ to 1.0,   PR 286 + PR310 + noise-supression-rebased. max_point_age_s_:8, min_num_points_per_cell_:12.

[...]

PR #325, opened 2019-04-25
Pointcloud noise suppression

nicovanduijn:
This is simply a rebase of #289, it should consider histogram cells free if they don't contain a minimum number of points, defined by the parameter `min_num_points_per_cell_`.

We will test-fly this soon. Attempting to solve #309 

@sunzj Can you test if this works for you?


sunzj:
@nicovanduijn as I said in https://github.com/PX4/avoidance/issues/309, the noise suppression is important for Realsense R200. 

Test outdoor, without noise suppression, the drone will stop even if there isn't any obstacle.

LorenzMeier:
@jkflying Please make sure we do suppress noise for all models, not just for the R200. Vision is never 100% accurate and I don't want to build a solution that deteriorates quickly in performance if the sensors are not perfect or the lighting is suboptimal.

nicovanduijn:
We need this capability for sure. The challenge will be finding a suppression level that works reasonably well for all models. I would suggest having the default values tuned to the most commonly used sensor, leaving individual tuning to each user

nicovanduijn:
We had successful flight tests with this.The default noise suppression level is set to a good value for SITL (1 pt/cell), since SITL is what most people use.

As to reasonable noise suppression levels on real flights, we have so far tested 25 pts/cell with an intel RealSense D435 and achieved pretty good results. We will test increasing it a bit more, but that should not be a blocker for this PR.

I added a sample config yaml file for real flights for convenience.

LGTM

PR #329, opened 2019-04-26
docker: Missing deps and outdated file path

potaito:
* For local-planner-node cmake would fail because of `libyaml-cpp` and `mavros-extras` missing. 
* The init files in px4 have been moved around / unified. There I'm not sure if the suggested file is the correct one.

### DIFF
--- a/docker/components/sitl-server/Dockerfile
+++ b/docker/components/sitl-server/Dockerfile
@@ -31,7 +31,7 @@ RUN ["/bin/bash", "-c", " \
         DONT_RUN=1 make posix_sitl_default gazebo \
 "]
 
-RUN echo "param set MAV_BROADCAST 1" >> ${FIRMWARE_DIR}/posix-configs/SITL/init/ekf2/iris
+RUN echo "param set MAV_BROADCAST 1" >> ${FIRMWARE_DIR}/ROMFS/px4fmu_common/init.d-posix/10016_iris}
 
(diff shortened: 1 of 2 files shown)


docker/components/sitl-server/Dockerfile
```diff
@@ -31,7 +31,7 @@ RUN ["/bin/bash", "-c", " \
         DONT_RUN=1 make posix_sitl_default gazebo \
 "]
 
-RUN echo "param set MAV_BROADCAST 1" >> ${FIRMWARE_DIR}/posix-configs/SITL/init/ekf2/iris
+RUN echo "param set MAV_BROADCAST 1" >> ${FIRMWARE_DIR}/ROMFS/px4fmu_common/init.d-posix/10016_iris}
```
mrivi: it could be `1015_iris_obs_avoid` such that `COM_OBS_AVOID` is enabled by default

potaito: Hm okay, I'll revisit this once I got it working in its entirety :)
Really not sure know...

potaito:
Let me figure out Martina's comment

nicovanduijn:
Doesn't this work: 
`RUN echo "param set MAV_BROADCAST 1" >> ${FIRMWARE_DIR}/ROMFS/px4fmu_common/init.d-posix/10015_iris_obs_avoid}`

PR #330, opened 2019-04-29
Simplify parsing

jkflying:
This removes a lot of the code duplication in the parsing, making it easier to read, verify and add additional fields.


jkflying:
Tested setting parameters in SITL, everything looks fine.

PR #331, opened 2019-04-30
Move loop out of constructor in global_planner

Jaeyoung-Lim:
A loop was being run from a constructor in the `path_handler_node`

This PR moves the loop into a `ros::Asyncspinner` which runs the pathhandler node at a fixed rate.

Tested in SITL

### DIFF
--- a/global_planner/src/nodes/path_handler_node.cpp
+++ b/global_planner/src/nodes/path_handler_node.cpp
@@ -2,8 +2,7 @@
 
 namespace global_planner {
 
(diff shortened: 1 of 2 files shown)


jkflying:
Have you SITL tested this? Looks good, much better than having an infinite loop in the constructor :+1:

Jaeyoung-Lim:
@jkflying Yes, I have tested this in SITL, behavior looks more or less the same :)

PR #334, opened 2019-05-01
Add common avoidance package for shared libraries

Jaeyoung-Lim:
This is a first step on having a common shared library for `global_planner` and `local_planner`

This moves the `common.cpp` in `local_planner` to a new package called `avoidance`, which both `global_planner` and `local_planner` depends on.

This will enable the two planners to have a common code base rather than having near identical codes in two places. Currently only the status enumerations are shared in  `avoidance/common.h`

It would be great if anyone can give suggestions on what should be shared and what should not be shared. I think things like rviz world loader being shared makes sense, as we can run both systems with the same simulated world examples

Just in case, I have tested this in SITL and everything seems to work.


nicovanduijn:
Awesome!
I guess longer term, something like the waypoint generator should be common. (Both planners have one!)
But they work very differently right now, so it will be a huge refactor. This is a great first step!

Jaeyoung-Lim:
@nicovandujin Exactly!

I think the next step is to make the `global_planner` first work with the rviz world loader first, so that we have a good idea of the differences of the two planners.

Regarding the wpgenerator, would it make sense to simply use the one from local planner and adapt global planner to it since global planner only works in off board mode?

nicovanduijn:
Global planner works in mission too

Jaeyoung-Lim:
Oh, wait...I missed that. I guess I will need to look more into how to make them one.

mrivi:
Really nice!

Jaeyoung-Lim:
@mrivi @nicovanduijn Any plans for getting this merged? I would like to work on the global planner based on a common code base rather than reimplementing it in global planner

jkflying:
We are low on test vehicles at the moment, hopefully we can test a bunch of things in the merge queue in the next few days.

Jaeyoung-Lim:
@jkflying I have tested this in SITL, and as far as I can see looks good!

baumanta:
Screenshot from 2019-05-07 15-25-26

There seems to be an error regarding mav_msgs when trying to build on balena, any ideas where that could be coming from @Jaeyoung-Lim ?

Jaeyoung-Lim:
@baumanta hmm...That is unexpected. Is it related to this PR?

mrivi:
@Jaeyoung-Lim  @baumanta - `mav_msgs` isn't used anywhere. We can remove the dependency altogether

Jaeyoung-Lim:
@mrivi I have got rid of mav_msgs @baumanta Can you retest this PR please?

jkflying:
It builds in Balena now, LGTM.

PR #335, opened 2019-05-01
Publish rvizworld even after rviz respawn

Jaeyoung-Lim:
This PR makes the rvizworldloader publish the marker array representing the world from the loop inside the rvizworldloader object, rather than in the `cmdloop` of the localplanner as currently done with `ifndef`

- This enables the world to be still visible even after rviz has been respawned
- This  gets rid of `startup_` state in `local_planner_node` which was needed only for initially publishing the world.
- No need to initialize the publishers manually
- Get rid of some `#ifdef` in the `cmdloop`


Jaeyoung-Lim:
@jkflying Yes, I have tested with `-DDISABLE_SIMULATION`

rebased and tested

PR #336, opened 2019-05-02
Remove auto exposure toggling

nicovanduijn:
This PR removes the `realsense_params.sh` script, which was used to toggle the auto exposure upon boot. This is no longer needed, as the realsense drivers now do this correctly.

Also removes the unused `tf_90_deg` transform as in #333 from the aero launchfile


nicovanduijn:
Correct, the namespace of that parameter has changed. If you wanted to keep this script around, it would need to be fixed

baumanta:
no need to keep the scripts if they are not necessary anymore

Issue #341, opened 2019-05-07
Add limitation for local_planner parameters


sunzj:
**Describe problem solved by the proposed feature**
After PR281 https://github.com/PX4/avoidance/pull/281, local planner use the parameter from PX4 firmware. 

However, there isn't limitation for these parameters, as i test, the MPC_XY_CRUISE get from firmware is 10, even i already set the speed to 1 in mission mode. The local planner use 10 as param_mpc_xy_cruise,thus cause avoid obstacle fail.

**Describe your preferred solution**
Check and add limitation for local_planner parameters, not just use firmware parameters.

jkflying:
Long term, we want the local_planner (and the global_planner) to be completely transparent when there are no obstacles. This means that it should fly at the speed given by the MPC_XY_CRUISE during mission mode. Only once there are obstacles does the planner slow down the vehicle, usually due to the planned path being outside of the FOV of the camera.

So, the correct solution is to reduce MPC_XY_CRUISE. Otherwise there is a different behavior when obstacle avoidance is running or not, even when there are not any obstacles. MPC_XY_CRUISE can be updated live during flight, for testing.

jiejiezhang:
@jkflying  From the PX4 Firmware code we can see:  The MPC_XY_CRUISE is Normal horizontal velocity in AUTO modes (includes Mission/RTL / hold / etc). But when we perform the mission task, we can modify the speed.

![mission](https://user-images.githubusercontent.com/23076781/57353185-45b67b00-719a-11e9-8ce1-6089c46ca87e.png)


In obstacle avoidance, we can't judge the speed by just using this parameter（MPC_XY_CRUISE）.

nicovanduijn:
I've gotten confused with this speed set by the mission item before. Back then, the speed set in the mission item is not being used by PX4. I just tested in SITL, and this seems to no longer be the case, so I guess you are correct

sunzj:
So the speed set in mission mode isn't MPC_XY_CRUISE, that's why i set the speed to 1 in mission mode, but still get 10 from PX4 firmware.

jkflying:
This makes sense, yes. We need to check the speed of the mission item and use that. It will be a little bit complicated since we need to find which messages they are from, but I agree it should be done to make it consistent with flying non-avoidance missions.

PR #342, opened 2019-05-08
check mission item speed

mrivi:
Fix for #341 .

### DIFF
--- a/local_planner/src/nodes/local_planner_node.cpp
+++ b/local_planner/src/nodes/local_planner_node.cpp
@@ -57,6 +57,9 @@ LocalPlannerNode::LocalPlannerNode(const ros::NodeHandle& nh,
       "/mavros/altitude", 1, &LocalPlannerNode::distanceSensorCallback, this);
   px4_param_sub_ = nh_.subscribe("/mavros/param/param_value", 1,
                                  &LocalPlannerNode::px4ParamsCallback, this);
+  mission_sub_ = nh_.subscribe("/mavros/mission/waypoints", 1,
+                               &LocalPlannerNode::missionCallback, this);
+
   mavros_vel_setpoint_pub_ = nh_.advertise<geometry_msgs::Twist>(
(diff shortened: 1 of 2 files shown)


local_planner/src/nodes/local_planner_node.cpp
```diff
@@ -536,13 +541,30 @@ void LocalPlannerNode::checkPx4Parameters() {
...
+    }
     request_param("MPC_COL_PREV_D", local_planner_->px4_.param_mpc_col_prev_d);
 
     std::this_thread::sleep_for(std::chrono::seconds(30));
   }
 }
 
+void LocalPlannerNode::missionCallback(const mavros_msgs::WaypointList& msg) {
+  for (auto waypoint : msg.waypoints) {
+    if (waypoint.command == 178 && (waypoint.param1 - 1.0f) < FLT_MIN &&
```
jkflying: Should it be `std::abs(waypoint.param1 - 1.0f) < FLT_MIN` ?

local_planner/src/nodes/local_planner_node.cpp
```diff
@@ -536,13 +541,30 @@ void LocalPlannerNode::checkPx4Parameters() {
...
+        waypoint.param2 > 0.0f) {
+      // MAV_CMD_DO_CHANGE_SPEED, speed type: ground speed, speed valid
+      local_planner_->px4_.param_mpc_xy_cruise = waypoint.param2;
```
jkflying: I'm not super happy with setting a named parameter with the value of a different parameter - it might lead to confusion later on. Maybe in our `px4_` we can add a `mission_custom_cruise` that is NaN when not set. Then when we set the speed we check first the custom cruise, and afterwards if it is NaN we take the `mpc_xy_cruise` as a fallback.

mrivi:
Closing in favor of https://github.com/PX4/avoidance/pull/434

PR #345, opened 2019-05-09
bug fix: enable drone to fly without smoothing

baumanta:
Setting smoothing velocity to zero should disable the smoothing. There was a bug in which the zero value was still used in the yaw smoothing which led to weird behavior.

Rename some variables to include units, such that they don't confuse me anymore

### DIFF
--- a/local_planner/src/nodes/waypoint_generator.cpp
+++ b/local_planner/src/nodes/waypoint_generator.cpp
@@ -111,7 +111,7 @@ void WaypointGenerator::updateState(const Eigen::Vector3f& act_pose,
   velocity_ = vel;
   goal_ = goal;
   prev_goal_ = prev_goal;
-  curr_yaw_ = getYawFromQuaternion(q) * DEG_TO_RAD;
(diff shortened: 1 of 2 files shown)


local_planner/src/nodes/waypoint_generator.cpp
```diff
@@ -211,25 +211,31 @@ void WaypointGenerator::smoothWaypoint(float dt) {
...
-  const float desired_yaw_velocity = 0.0;
+  if (smoothing_speed_xy_ > 0) {
+    const float P_constant_xy = smoothing_speed_xy_;
+    const float D_constant_xy =
+        2.f * std::sqrt(P_constant_xy);  // critically damped
 
-  float yaw_diff = std::isfinite(desired_setpoint_yaw)
-                       ? desired_setpoint_yaw - setpoint_yaw_
-                       : 0.0f;
+    const float desired_setpoint_yaw_rad =
```
jkflying: I think move this calculation outside of the if, then in the else you can just set it directly.

PR #349, opened 2019-05-10
Only set yaw when goal is specific distance away

nicovanduijn:
This fixes the bug that take-off was always done facing east

### DIFF
--- a/local_planner/src/nodes/waypoint_generator.cpp
+++ b/local_planner/src/nodes/waypoint_generator.cpp
@@ -214,14 +214,21 @@ void WaypointGenerator::nextSmoothYaw(float dt) {
   // and we dont want to introduce yet another parameter
 
   const float desired_setpoint_yaw_rad =
-      nextYaw(position_, output_.goto_position);
+      (position_.topRows<2>() - output_.goto_position.topRows<2>()).norm() >
+              0.1f
+          ? nextYaw(position_, output_.goto_position)
+          : curr_yaw_rad_;
 
-  if (smoothing_speed_xy_ > 0) {
+  // If smoothing is disabled, set yaw to face goal directly
+  if (smoothing_speed_xy_ <= 0.01f) {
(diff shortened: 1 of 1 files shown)


local_planner/src/nodes/waypoint_generator.cpp
```diff
@@ -214,14 +216,20 @@ void WaypointGenerator::nextSmoothYaw(float dt) {
...
-      nextYaw(position_, output_.goto_position);
+      (position_ - output_.goto_position).normXY() > 0.1f
+          ? nextYaw(position_, output_.goto_position)
+          : curr_yaw_rad_;
 
-  if (smoothing_speed_xy_ > 0) {
+  // If smoothing is disabled, set yaw to face goal directly
+  if (smoothing_speed_xy_ <= 0.01f) {
+    setpoint_yaw_rad_ = desired_setpoint_yaw_rad;
+    return;
```
jkflying: I don't think you need the return here, since everything is inside of the `else` afterwards

local_planner/src/nodes/waypoint_generator.cpp
```diff
@@ -214,14 +216,20 @@ void WaypointGenerator::nextSmoothYaw(float dt) {
   // and we dont want to introduce yet another parameter
 
   const float desired_setpoint_yaw_rad =
-      nextYaw(position_, output_.goto_position);
+      (position_ - output_.goto_position).normXY() > 0.1f
+          ? nextYaw(position_, output_.goto_position)
+          : curr_yaw_rad_;
```
jkflying: Nice!

local_planner/src/nodes/waypoint_generator.cpp
nicovanduijn: Good point, my idea behind this refactor was reducing the nested if() clauses with an early exit. Similar to SmoothWaypoint(). I'll remove the else

PR #350, opened 2019-05-11
Get rid of redundant state planner_is_healthy

Jaeyoung-Lim:
Currently the status of the planner is monitored by two states
- `planner_is_healthy`
- `never_run_`

This PR gets rid of redundant state `planner_is_healthy`

As far as I can see, with the current logic there is no additional information we are getting by using `planner_is_healthy` as a state

This also makes all `setSystemSatus` calls to be inside the `checkFailsafe` function.

### DIFF
--- a/local_planner/src/nodes/local_planner_node.cpp
+++ b/local_planner/src/nodes/local_planner_node.cpp
@@ -337,13 +337,6 @@ void LocalPlannerNode::cmdLoopCallback(const ros::TimerEvent& event) {
   if (companion_state_ == MAV_STATE::MAV_STATE_ACTIVE)
     calculateWaypoints(hover_);
 
-  if (!never_run_) {
-    for (size_t i = 0; i < cameras_.size(); ++i) {
-      // once the camera info have been set once, unsubscribe from topic
-      cameras_[i].camera_info_sub_.shutdown();
-    }
-  }
-
   position_received_ = false;
 
   // publish system status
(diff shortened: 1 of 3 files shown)


local_planner/src/nodes/local_planner_node.cpp
```diff
@@ -338,16 +337,16 @@ void LocalPlannerNode::cmdLoopCallback(const ros::TimerEvent& event) {
...
   updatePlanner();
 
   // send waypoint
-  if (!never_run_ && planner_is_healthy_) {
+  if (companion_state_ == MAV_STATE::MAV_STATE_ACTIVE)
     calculateWaypoints(hover_);
-    if (!hover_) setSystemStatus(MAV_STATE::MAV_STATE_ACTIVE);
-  } else {
+
+  if (!never_run_) {
```
baumanta: do we need to unsubscribe from the camera_info topic in every iteration?

Jaeyoung-Lim: No, you are right. I will try and fix this

Jaeyoung-Lim:
@baumanta 

I went on calling `shutdown` inside the callback function. I have verified that indeed the camera info is only read once for each of the camera topics.

This also makes `never_run_` no longer necessary, so I went on to get rid of it.

Jaeyoung-Lim:
CI failing is unrelated to the build, could anyone of the admins restart the build please?

Jaeyoung-Lim:
@baumanta I am really not sure.

As far as I can see, I think a flight test is not necessary. The changes in the logic should be all captured in SITL. However, I think you will have much more insight whether if this change will affect real flight.

Jaeyoung-Lim:
@jkflying Rebased and tested in SITL!

PR #352, opened 2019-05-14
Fix Deadzone FOV bug

nicovanduijn:
This commit fixes a bug observed in real flight:
Due to overestimation of the actual FOV of the point cloud, points were never remembered because they disappeard in the blind spot.

This commit replaces the inside-FOV detection in the point cloud processing to use the max and min horizontal angle of the incoming points.

### DIFF
--- a/local_planner/src/nodes/planner_functions.cpp
+++ b/local_planner/src/nodes/planner_functions.cpp
@@ -22,7 +22,7 @@ void processPointcloud(
   final_cloud.width = 0;
   final_cloud.points.reserve((2 * GRID_LENGTH_Z) * (2 * GRID_LENGTH_E));
 
-  float  distance = 0.f;
+  float distance = 0.f;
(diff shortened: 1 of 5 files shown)


nicovanduijn:
Lots of stuff wrong with this - e.g. when there is no point cloud we would accumulate noise points.

Closing in favor of #358

local_planner/src/nodes/planner_functions.cpp
```diff
@@ -39,6 +40,8 @@ void processPointcloud(
               distance < histogram_box.radius_) {
             // subsampling the cloud
             PolarPoint p_pol = cartesianToPolar(toEigen(xyz), position);
+            h_min = std::min(h_min, p_pol.z);
```
baumanta: I like that solution :) maybe we could use a bit more descriptive names such as z_FOV_angle_min or similar?

Issue #357, opened 2019-05-15
Memory not working in real flight due to blind spot


nicovanduijn:
During real flight, the local planner does not remember previously seen points.

The issue is that we are overestimating the FOV of the sensor slightly, yielding a blind spot at the edge of the sensor. As points move along, they disappear from the point cloud before we consider them outside the FOV - hence they never enter the local point cloud memory

mrivi:
Fixed

PR #358, opened 2019-05-15
Remove FOV indices

nicovanduijn:
This is a refactor of the way the FOV is kept internally - with current orientation and sensor FOV instead of histogram indices.

The end goal of this refactor is to address #357 

This has the (theoretical) drawback that we could not account for discontinuous FOV anymore (multiple cameras...). This hasn't been done so far anyway

### DIFF
--- a/local_planner/test/test_local_planner.cpp
+++ b/local_planner/test/test_local_planner.cpp
@@ -5,6 +5,9 @@
 #include "../include/local_planner/local_planner.h"
 #include "avoidance/common.h"
 
+#define TO_DEG 180.f / M_PI_F
+#define TO_RAD M_PI_F / 180.f
+
 // Stateless tests:
 // Create some hardcoded scan data of obstacles in different positions
 // For each one check that the planner response is correct
@@ -21,6 +24,7 @@ class LocalPlannerTests : public ::testing::Test {
     avoidance::LocalPlannerNodeConfig config =
         avoidance::LocalPlannerNodeConfig::__getDefault__();
     planner.dynamicReconfigureSetParams(config, 1);
+    planner.setFOV(59.0f, 46.0f);
 
     // start with basic pose
     Eigen::Vector3f pos(0.f, 0.f, 0.f);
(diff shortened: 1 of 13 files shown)


local_planner/test/test_local_planner.cpp
```diff
@@ -85,9 +88,14 @@ TEST_F(LocalPlannerTests, all_obstacles) {
   // THEN: it should get a scan showing the obstacle
   sensor_msgs::LaserScan scan;
   planner.getObstacleDistanceData(scan);
+  int idx_lower_obstacle_boundary_bin = static_cast<int>(
```
jkflying: This is made possible by using the entire pointcloud for the laser scan, and not limiting it to FOV, right?

local_planner/test/test_planner_functions.cpp
```diff
@@ -174,40 +108,43 @@ TEST(PlannerFunctionsTests, processPointcloud) {
...
-  FOV_zero.z_idx_vec.push_back(0);
-
-  FOV_indices FOV_regular;
-  FOV_regular.e_idx_max = 10;
-  FOV_regular.e_idx_min = 6;
-  FOV_regular.z_idx_vec.push_back(21);
+  processed_cloud1.push_back(toXYZI(position + memory_point, 5.0f));
+  processed_cloud2.push_back(toXYZI(position + memory_point, 5.0f));
+  processed_cloud3.push_back(toXYZI(position + memory_point, 5.0f));
+  std::cout << "mem point e: " << memory_point_polar.e
```
jkflying: remove the `cout`

local_planner/test/test_local_planner.cpp
nicovanduijn: Actually it would have been possible before:
```
int idx_upper_obstacle_boundary_bin = static_cast<int>(
      (M_PI_F / 2 - atan2(fov_half_y, distance)) / (scan.angle_increment));
```

local_planner/test/test_planner_functions.cpp
nicovanduijn: ha. missed that one

local_planner/test/test_local_planner.cpp
jkflying: I think you'd have to take a `min()`/`max()` of the FOV and the actual obstacle, depending on the case.

baumanta:
looks good, I also tested that in SITL and it works with the new collision prevention draft. The vehicle cannot go towards obstacles outside the FOV which were remembered

PR #359, opened 2019-05-15
Test avoidance pkg

### LINKED ISSUE #356: Common tests don't compile and aren't run in CI
Currently, the tests in `avoidance/test` don't compile and are not run in CI

To reproduce:
```
cd avoidance
catkin run_tests --this --no-deps
```

Output:
image

I believe the CMakeLists.txt need to include histogram.h, but I don't want to introduce a dependency of common to the local planner. (Which would yield circular dependency)

Not sure why these tests need `histogram.h` anyway

@Jaeyoung-Lim Any ideas?

Jaeyoung-Lim:
This PR enables / fixes tests for the `avoidance` package.

- moved histogram to avoidance pkg
- fixed Travis to also test avoidance

This fixes #356


Jaeyoung-Lim:
Wow, I was missing out a lot of things...Thanks!

PR #360, opened 2019-05-15
Remove repeated info outputs in visualizer

Jaeyoung-Lim:
After visualizing the world moved to `WorldVisualizer`, `ROSINFO` is flooding the console output with `Successfully loaded rviz world` as below
```
[ INFO] [1557945199.308421401, 20.588000000]: Creating 1 swatches
[ INFO] [1557945200.522503617, 21.600000000]: Creating 1 swatches
[ INFO] [1557945200.961392230, 22.000000000]: Successfully loaded rviz world
[ INFO] [1557945201.408963923, 22.436000000]: HP: requesting home position
[ INFO] [1557945201.611341746, 22.636000000]: Creating 1 swatches
[ INFO] [1557945202.603751584, 23.612000000]: Creating 1 swatches
[ INFO] [1557945203.004797869, 24.004000000]: Successfully loaded rviz world
[ INFO] [1557945203.467378202, 24.452000000]: Creating 1 swatches
[ INFO] [1557945204.459814082, 25.212000000]: Creating 1 swatches
[ INFO] [1557945205.228142521, 25.920000000]: Creating 1 swatches
[ INFO] [1557945205.317867858, 26.004000000]: Successfully loaded rviz world
```
This PR makes the statement be printed only once

### DIFF
--- a/avoidance/src/rviz_world_loader.cpp
+++ b/avoidance/src/rviz_world_loader.cpp
@@ -120,7 +120,7 @@ int WorldVisualizer::visualizeRVIZWorld(const std::string& world_path) {
   }
 
   world_pub_.publish(marker_array);
-  ROS_INFO("Successfully loaded rviz world");
+  ROS_INFO_ONCE("Successfully loaded rviz world");
   return 0;
 }
 


jkflying:
Is this an indication that it was only being published once before, and now is happening once per second? Is there any potential for performance problems here, eg. is rviz going to run out of memory?

Jaeyoung-Lim:
@jkflying This was introduced after #335 as the rviz world was published repeatedly(Once per second), to be able to visualize the world even after `rviz` is closed / reopened.->This was my intention.

It happened to me that sometimes when I open rviz separately, the rviz world never makes it to rviz.

I think it would be worth looking into if rviz stores the history of the published worlds, but I doubt that this will lead to high memory usage on the system. 

Also, we can also lower the world publishing rate further more.

mhkabir:
What if you just make the visualisations a latched topic - this way you don't need to repeatedly call publish. 

New subscribers always gets the last message that was published, even if the message was published before the subscriber was connected.

Jaeyoung-Lim:
@mhkabir I think previously it was done that way, but in that case it will still not be able to visualize the world if rviz is respawned

Jaeyoung-Lim:
@jkflying I checked the bandwidth of how much bandwidth the world publishing consumes at (0.5Hz), and it is ~1.5KB/s.

Would we want to go back to publishing the world once?

nicovanduijn:
Nice, I wasn't even aware of the `ROS_INFO_ONCE` API. I'm fine with this

PR #362, opened 2019-05-16
Deprecate threePointMsg in global_planner

### LINKED ISSUE #354: Quesntion on ThreePointMsg.msg support
`ThreePointMsg.msg` used in `global_planner/path_handler_node` seems to be doing something [quite similar](https://github.com/PX4/avoidance/blob/master/global_planner/src/nodes/path_handler_node.cpp#L312
) to what `mavros_msgs/trajectory` is sending. 

Would it make sense to depricate `ThreePointMsg` in global planner? or is there a reason that this is needed?

Jaeyoung-Lim:
This PR deprecates the threePointMsg custom message implementation in `global_planner`.

Initially it was quite confusing for me to understand why this message exists in the first place. It seems like it was a interface that was a precursor of the Avoidance triplet. This Fixes #354 

This does not affect the performance of global_planner.

Tested in SITL, both in offboard and mission

### DIFF
--- a/global_planner/CMakeLists.txt
+++ b/global_planner/CMakeLists.txt
@@ -64,7 +64,6 @@ find_package(octomap REQUIRED)
 add_message_files(
   FILES
   PathWithRiskMsg.msg
-  ThreePointMsg.msg
 )
 
 ## Generate services in the 'srv' folder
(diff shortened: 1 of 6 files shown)


nicovanduijn:
-122 lines of code :1st_place_medal:

PR #364, opened 2019-05-16
Dynamically estimate FOV based on point cloud

nicovanduijn:
This PR works on top of #358 and computes the FOV dynamically based on the point cloud.

At the same time, an error margin is subtracted in order to always underestimate it. This is needed because delay in the system (current yaw estimation and most recent point cloud during fast yaw maneuvers) can yield to overestimating the FOV.

Overestimating the FOV, even just a few degrees leads to the memory bug described in #357

### DIFF
--- a/avoidance/src/common.cpp
+++ b/avoidance/src/common.cpp
@@ -11,10 +11,10 @@ namespace avoidance {
 bool pointInsideFOV(const FOV& fov, const PolarPoint& p_pol) {
   // pitch angle must be counted negative: zero pitch is level, positive pitch
   // is leaning down. Whereas a positive elevation angle is counted upwards
-  return p_pol.z <= fov.yaw_deg + fov.h_fov_deg / 2.f &&
-         p_pol.z >= fov.yaw_deg - fov.h_fov_deg / 2.f &&
-         p_pol.e <= -fov.pitch_deg + fov.v_fov_deg / 2.f &&
-         p_pol.e >= -fov.pitch_deg - fov.v_fov_deg / 2.f;
+  return p_pol.z <= fov.azimuth_deg + fov.h_fov_deg / 2.f &&
+         p_pol.z >= fov.azimuth_deg - fov.h_fov_deg / 2.f &&
+         p_pol.e <= -fov.elevation_deg + fov.v_fov_deg / 2.f &&
+         p_pol.e >= -fov.elevation_deg - fov.v_fov_deg / 2.f;
 }
 
 float distance2DPolar(const PolarPoint& p1, const PolarPoint& p2) {
--- a/avoidance/src/common.cpp
+++ b/avoidance/src/common.cpp
@@ -9,12 +9,12 @@
 namespace avoidance {
 
 bool pointInsideFOV(const FOV& fov, const PolarPoint& p_pol) {
-  // pitch angle must be counted negative: zero pitch is level, positive pitch
-  // is leaning down. Whereas a positive elevation angle is counted upwards
-  return p_pol.z <= fov.azimuth_deg + fov.h_fov_deg / 2.f &&
-         p_pol.z >= fov.azimuth_deg - fov.h_fov_deg / 2.f &&
-         p_pol.e <= -fov.elevation_deg + fov.v_fov_deg / 2.f &&
-         p_pol.e >= -fov.elevation_deg - fov.v_fov_deg / 2.f;
+  return p_pol.z <=
+             wrapAngleToPlusMinus180(fov.azimuth_deg + fov.h_fov_deg / 2.f) &&
+         p_pol.z >=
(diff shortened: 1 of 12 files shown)


local_planner/src/nodes/local_planner.cpp
```diff
@@ -16,11 +16,10 @@ LocalPlanner::~LocalPlanner() {}
 void LocalPlanner::setPose(const Eigen::Vector3f& pos,
                            const Eigen::Quaternionf& q) {
   position_ = pos;
-  curr_yaw_fcu_frame_deg_ = getYawFromQuaternion(q);
-  curr_yaw_histogram_frame_deg_ = -curr_yaw_fcu_frame_deg_ + 90.0f;
-
-  curr_pitch_fcu_frame_deg_ = getPitchFromQuaternion(q);
-  star_planner_->setPose(position_, curr_yaw_histogram_frame_deg_);
+  fov_.azimuth_deg = -getYawFromQuaternion(q) + 90.0f;
+  wrapAngleToPlusMinus180(fov_.azimuth_deg);
```
baumanta: could we do that in one line? e.g 
`fov_.azimuth_deg = wrapAngleToPlusMinus180(-getYawFromQuaternion(q) + 90.0f);`

and maybe put a comment above that those lines transform the yaw and pitch angles in fcu frame to the histogram convention

local_planner/src/nodes/local_planner_visualization.cpp
```diff
@@ -93,14 +93,14 @@ void LocalPlannerVisualization::publishFOV(const Eigen::Vector3f& drone_pos,
                                            const float max_range) const {
   // pitch angle must be counted negative: zero pitch is level, positive pitch
```
baumanta: I think you can take out this comment as it is not necessary anymore

avoidance/src/common.cpp
```diff
@@ -11,10 +11,10 @@ namespace avoidance {
...
   // pitch angle must be counted negative: zero pitch is level, positive pitch
   // is leaning down. Whereas a positive elevation angle is counted upwards
-  return p_pol.z <= fov.yaw_deg + fov.h_fov_deg / 2.f &&
-         p_pol.z >= fov.yaw_deg - fov.h_fov_deg / 2.f &&
-         p_pol.e <= -fov.pitch_deg + fov.v_fov_deg / 2.f &&
-         p_pol.e >= -fov.pitch_deg - fov.v_fov_deg / 2.f;
+  return p_pol.z <= fov.azimuth_deg + fov.h_fov_deg / 2.f &&
+         p_pol.z >= fov.azimuth_deg - fov.h_fov_deg / 2.f &&
+         p_pol.e <= -fov.elevation_deg + fov.v_fov_deg / 2.f &&
+         p_pol.e >= -fov.elevation_deg - fov.v_fov_deg / 2.f;
```
baumanta: I think here the minus is wrong. Now that we use elevation angles, we can use them without the minus

local_planner/src/nodes/planner_functions.cpp
```diff
@@ -24,6 +24,13 @@ void processPointcloud(
 
   float distance;
 
+  float fov_yaw_lower_boundary = fov.azimuth_deg - fov.h_fov_deg / 2.0f;
```
baumanta: maybe just to stay consistent, rename the variables here to `fov_azimuth_lower_boundary`, `fov_elevation_lower_boundary`, ...

local_planner/src/nodes/local_planner.cpp
nicovanduijn: nope, you can't modify an rvalue ;)

I can add the comments

baumanta: ok fine leave it the was it is, I did not look at function close enough. I just looked a bit weird to me :)

jkflying: I'd prefer if we could get these returning values, rather than doing it silently on references. It allows chaining functions for exactly things like his, making a lot of otherwise complex code easy to do, and it makes it easier to reformulate code into 'pipelines'.

avoidance/src/common.cpp
```diff
@@ -142,16 +142,16 @@ float getPitchFromQuaternion(const Eigen::Quaternionf q) {
   return pitch * RAD_TO_DEG;
 }
 
-void wrapAngleToPlusMinusPI(float& angle) {
-  angle = angle - 2.0f * M_PI_F * std::floor(angle / (2.0f * M_PI_F) + 0.5f);
+float wrapAngleToPlusMinusPI(float angle) {
```
jkflying: If we're going to change these, then we should add a 
```c++
#define WARN_UNUSED __attribute__((warn_unused_result))
```
```c++
float WARN_UNUSED wrapAngleToPlusMinusPI(float angle) {
```
In case anybody uses the old convention by mistake. And then we

PR #366, opened 2019-05-17
Add cmake flags for compilation with disabled sim

nicovanduijn:
This commit adds a check in the cmake files for the DISABLE_SIMULATION
flag, which was previously broken

Output when compiling with
```
catkin config --cmake-args -DCMAKE_BUILD_TYPE=Release -DCATKIN_ENABLE_TESTING=False -DDISABLE_SIMULATION=ON
catkin build
```

image

@Jaeyoung-Lim Are you able to build with `-DDISABLE_SIMULATION` without this fix?


Jaeyoung-Lim:
@nicovanduijn It seems like it was a regression from #359, which the changes were made by @jkflying to fix a yaml-cpp dependency

PR #368, opened 2019-05-17
deprecate docker

mrivi:
This PR removes the docker containers from the avoidance repository. This project is undergoing fast development and the dev team is currently not able to keep maintaining these containers. 
Furthermore the avoidance CI is running on the PX4 official containers, so we should focus on keep maintaining those.


potaito:
Wouldn't it be enough to rename `docker` to `docker_deprecated` ?

jkflying:
There is also the `.travis.yml` which shows how to install it.

PR #369, opened 2019-05-20
Set default PX4 parameters at first.

sunzj:
1.To avoid read param fail cause issue,
  set default px4 parameters at first.


jkflying:
We've seen this issue too. We need to be careful with loading the defaults though, because it might lead to unsafe behavior until the real values are loaded. We need to investigate why this takes so long - it should take less than 100ms for all of the data to be loaded. I guess there are dropped messages or something similar.

jkflying:
This is fixed by #421

PR #370, opened 2019-05-20
Add down-sampled depth images for logging

nicovanduijn:
This commit adds a down-sampled version of the depth images of each camera to the auto-generated launch file.

This results in an additional topic `CAMERA_NAME/depth/image_rect_raw_drop`, published at 1/15th the rate of the original topic. 

For our current rates on this topic (15Hz) this results in a bandwidth of 650kb/s for each camera, which is a reasonable size to log, even on embedded platforms.


nicovanduijn:
Ideally, we would do this with the infra red images, but I cannot get the realsense driver to publish that stream, regardless of what configuration I set. I believe it may be a USB2.0 vs 3.0 issue:
image

nicovanduijn:
Disabling the second infra stream makes the first one work - this seems to be a limitation of using USB2.1 vs. USB3.1
Added downsampled IR image as well

nicovanduijn:
I think this needs some further investigation before merging. Reason being that it moves the point cloud generation from the ROS wrapper into the driver.
At first sight, this seemed to be more efficient (I was running just the driver and subscribing to the cloud)

Somehow I'm now getting increased CPU load when the planner is running with 3 cameras:
image

I find it also suspiscious that one camera takes ~50% CPU while the others are round 40%. But in turn the planner seems to consume less.

I'll investigate a bit more

jkflying:
There's a small conflict now with the Structure Core, since that publishes the point cloud at /sc/depth/points and you've now changed it to /$1/depth/color/points. So that either needs to go back to what it was, or we need to use separate topic names depending on the camera type.

Also, I saw in the ROS wrapper that the colored point cloud might be cropped to fit the RGB camera FOV. There is now an option to turn this off: https://github.com/IntelRealSense/realsense-ros/commit/b4060b618c0dccead44dad09ecfd6f19f766c2b1

PR #372, opened 2019-05-21
Fix FOV pitch angle

baumanta:
The sign of the pitch angle was wrong in the calculation of the FOV. This led to the FOV shifting upwards if the drone was tilting down.

I added the FOV visualization which I used to find the bug, as I think it would be generally useful


nicovanduijn:
Can you add a test such that the old bug would be caught? (E.g. a point that is outside the FOV due to current pitch and vertical FOV)

PR #374, opened 2019-05-22
Plan to sensor horizon at all times

nicovanduijn:
This commit removes the 'max_tree_length' parameter, replacing its
logic with the box_size. This way we always plan to the maximum
of the sensor range, which is a logical thing to do.

To throttle the planning effort the number of nodes expanded during
A* search is much more suitable, and that logic remains untouched.


baumanta:
I would really love if that works, I did not really like that we needed to constrain path length that much. But I think this needs to be tested in SITL and flight tests to see if it is truly ok now. When I added that constrain, it was necessary to avoid big detours.

nicovanduijn:
For what it's worth, it works pretty well in SITL.

It does fly farther around things, yes. But I kind of like it

Jaeyoung-Lim:
@nicovanduijn Does this bring any performance gains e.g. maximum speed?

PR #384, opened 2019-05-29
Add robot radius for path searching in global_planner

Jaeyoung-Lim:
Prior to this PR, the path found from the global_planner would tend to plan paths that are too close to the obstacle, or the paths would go through extremely narrow paths which the drone will not be able to traverse.

This PR adds a robot radius to consider when calculating the list of neighbors in global planner.

The radius can be defined from a parameter as the following in the `global_planner_node`
```
    <node name="global_planner_node" pkg="global_planner" type="global_planner_node" output="screen"
       ...
        <param name="robot_radius" value="1.0" />        
    </node>
```

Tested in SITL, and the behavior seems reasonable (plans around further away with larger robot_radius)

### DIFF
--- a/global_planner/src/library/cell.cpp
+++ b/global_planner/src/library/cell.cpp
@@ -72,11 +72,12 @@ Cell Cell::getNeighborFromYaw(double yaw) const {
 // Returns the neighbors of the Cell whose risk influences the Cell
 std::vector<Cell> Cell::getFlowNeighbors(int radius) const {
   std::vector<Cell> cells;
-  for(int x = -radius; x < radius; x++){
-    for(int y = -radius; x < radius; x++){
-      for(int z = -radius; x < radius; x++){
-        cells.push_back(Cell(std::tuple<int, int, int>(xIndex() + x, yIndex() + y, zIndex() + z)));
-      }   
+  for (int x = -radius; x < radius; x++) {
+    for (int y = -radius; x < radius; x++) {
+      for (int z = -radius; x < radius; x++) {
+        cells.push_back(Cell(std::tuple<int, int, int>(
+            xIndex() + x, yIndex() + y, zIndex() + z)));
+      }
     }
   }
   return cells;
--- a/global_planner/src/library/cell.cpp
+++ b/global_planner/src/library/cell.cpp
@@ -73,9 +73,11 @@ Cell Cell::getNeighborFromYaw(double yaw) const {
 std::vector<Cell> Cell::getFlowNeighbors(int radius) const {
   std::vector<Cell> cells;
   for (int x = -radius; x <= radius; x++) {
-    int y_radius = static_cast<int>(std::ceil(std::sqrt(std::pow(double(radius), 2) - std::pow(double(x),2))));
+    int y_radius = static_cast<int>(std::ceil(
+        std::sqrt(std::pow(double(radius), 2) - std::pow(double(x), 2))));
     for (int y = -y_radius; y <= y_radius; y++) {
-      int z_radius = static_cast<int>(std::ceil(std::sqrt(std::pow(double(y_radius), 2) - std::pow(double(y),2))));
+      int z_radius = static_cast<int>(std::ceil(
+          std::sqrt(std::pow(double(y_radius), 2) - std::pow(double(y), 2))));
       for (int z = -z_radius; z <= z_radius; z++) {
         cells.push_back(Cell(std::tuple<int, int, int>(
             xIndex() + x, yIndex() + y, zIndex() + z)));
--- a/global_planner/src/library/cell.cpp
+++ b/global_planner/src/library/cell.cpp
@@ -73,8 +73,10 @@ Cell Cell::getNeighborFromYaw(double yaw) const {
 std::vector<Cell> Cell::getFlowNeighbors(int radius) const {
   std::vector<Cell> cells;
   for (int x = -radius; x <= radius; x++) {
-    for (int y = -radius; y <= radius; y++) {
-      for (int z = -radius; z <= radius; z++) {
+    int y_radius = static_cast<int>(std::ceil(std::sqrt(std::pow(double(radius), 2) - std::pow(double(x),2))));
+    for (int y = -y_radius; y <= y_radius; y++) {
+      int z_radius = static_cast<int>(std::ceil(std::sqrt(std::pow(double(y_radius), 2) - std::pow(double(y),2))));
+      for (int z = -z_radius; z <= z_radius; z++) {
         cells.push_back(Cell(std::tuple<int, int, int>(
             xIndex() + x, yIndex() + y, zIndex() + z)));
       }
--- a/global_planner/src/library/cell.cpp
+++ b/global_planner/src/library/cell.cpp
@@ -72,14 +72,15 @@ Cell Cell::getNeighborFromYaw(double yaw) const {
 // Returns the neighbors of the Cell whose risk influences the Cell
 std::vector<Cell> Cell::getFlowNeighbors(int radius) const {
   std::vector<Cell> cells;
-  auto ceilDistance = [](int radius, int x) {
-    return static_cast<int>(std::ceil(
-        std::sqrt(std::pow(double(radius), 2) - std::pow(double(x), 2))));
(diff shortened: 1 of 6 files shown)


global_planner/src/library/global_planner.cpp
```diff
@@ -191,7 +192,8 @@ double GlobalPlanner::getRisk(const Cell& cell) {
   }
 
   double risk = getSingleCellRisk(cell);
-  for (const Cell& neighbor : cell.getFlowNeighbors()) {
+  int radius = int(robot_radius_ / octree_resolution_);
```
jkflying: How about 
```suggestion
  int radius = static_cast<int>(std::ceil(robot_radius_ / octree_resolution_));
```

Since the octree resolution can be very course, so it might round to 0.

Jaeyoung-Lim: Looks much better!

global_planner/src/library/cell.cpp
```diff
@@ -70,14 +70,17 @@ Cell Cell::getNeighborFromYaw(double yaw) const {
...
-  return std::vector<Cell>{
-      Cell(std::tuple<int, int, int>(xIndex() + 1, yIndex(), zIndex())),
-      Cell(std::tuple<int, int, int>(xIndex() - 1, yIndex(), zIndex())),
-      Cell(std::tuple<int, int, int>(xIndex(), yIndex() + 1, zIndex())),
-      Cell(std::tuple<int, int, int>(xIndex(), yIndex() - 1, zIndex())),
-      Cell(std::tuple<int, int, int>(xIndex(), yIndex(), zIndex() + 1)),
-      Cell(std::tuple<int, int, int>(xIndex(), yIndex(), zIndex() - 1))};
+std::vector<Cell> Cell::getFlowNeighbors(int radius) const {
+  std::vector<Cell> cells;
+  for (int x = -radius; x < radius; x++) {
```
SwiftGust: ``
```suggestion
  for (int x = -radius; x <= radius; x++) {
```
Merely a suggestion
I think should this line should include the equal
In case of 
robot_radius_ = 1;
octomap_resolution_ = 1;

only considers index of -1, 0

global_planner/src/library/cell.cpp
```diff
@@ -70,14 +70,17 @@ Cell Cell::getNeighborFromYaw(double yaw) const {
...
+    for (int y = -radius; x < radius; x++) {
```
SwiftGust:
```suggestion
    for (int y = -radius; y <= radius; y++) {
```
Is this typo? only checking & increasing x, not y

global_planner/src/library/cell.cpp
```diff
@@ -70,14 +70,17 @@ Cell Cell::getNeighborFromYaw(double yaw) const {
...
+      for (int z = -radius; x < radius; x++) {
```
SwiftGust:
```suggestion
      for (int z = -radius; z <= radius; z++) {
```
Same here, only checking & increasing x, not z

global_planner/src/nodes/global_planner_node.cpp
```diff
@@ -103,6 +103,9 @@ void GlobalPlannerNode::readParams() {
   nh_.param<double>("start_pos_z", start_pos_.z, 3.5);
   global_planner_.goal_pos_ =
       GoalCell(start_pos_.x, start_pos_.y, start_pos_.z);
+  double robot_radius;
+  nh_.param<double>("robot_radius", robot_radius, 0.5);
```
SwiftGust:
```suggestion
  nh_.param<double>("robot_radius", robot_radius, 1.0);
```
For the consistency with global_planner_octomap.launch file

Jaeyoung-Lim:
@SwiftGust Good catch! I have applied the suggestions

Jaeyoung-Lim:
Limited the considered neighbors to a sphere

global_planner/src/library/cell.cpp
```diff
@@ -70,14 +70,21 @@ Cell Cell::getNeighborFromYaw(double yaw) const {
...
-      Cell(std::tuple<int, int, int>(xIndex(), yIndex(), zIndex() + 1)),
-      Cell(std::tuple<int, int, int>(xIndex(), yIndex(), zIndex() - 1))};
+std::vector<Cell> Cell::getFlowNeighbors(int radius) const {
+  std::vector<Cell> cells;
+  for (int x = -radius; x <= radius; x++) {
+    int y_radius = static_cast<int>(std::ceil(
+        std::sqrt(std::pow(double(radius), 2) - std::pow(double(x), 2))));
+    for (int y = -y_radius; y <= y_radius; y++) {
+      int z_radius = static_cast<int>(std::ceil(
+          std::sqrt(std::pow(double(y_radius), 2) - std::pow(double(y), 2))));
```
jkflying: I think pull out the whole `static_cast<int>(std::ceil(std::sqrt(std::pow(double(radius), 2) - std::pow(double(x), 2))))` into a lambda something like `ceilDistance(radius, x) --> int` will make it a lot nicer to read

Jaeyoung-Lim: I have added the lambda in the following commit! This is cool!

global_planner/src/library/cell.cpp
```diff
@@ -70,14 +70,23 @@ Cell Cell::getNeighborFromYaw(double yaw) const {
...
-      Cell(std::tuple<int, int, int>(xIndex(), yIndex() - 1, zIndex())),
-      Cell(std::tuple<int, int, int>(xIndex(), yIndex(), zIndex() + 1)),
-      Cell(std::tuple<int, int, int>(xIndex(), yIndex(), zIndex() - 1))};
+std::vector<Cell> Cell::getFlowNeighbors(int radius) const {
+  std::vector<Cell> cells;
+  auto ceilDistance = [](int radius, int x) {
+    return static_cast<int>(std::ceil(
+        std::sqrt(std::pow(double(radius), 2) - std::pow(double(x), 2))));
+  };
+  for (int x = -radius; x <= radius; x++) {
```
jkflying: Much easier to read. However, I think it isn't actually getting the radius we expect, since the z radius only knows about the y and not the x.

The naive way to do it something like this, generate a cube and filter out the non-sphere parts:
```c++
auto inSphere = [radius](int x, int y, int z,) -> bool {
    return std::ceil(std::sqrt(x*x + y*y + z*z)) >= radius;
};

for (int x = -radius; x <= radius; x++) {
    for (int y = -radius; y <= radius; y++) {
        for(int z = -radius; z <= radius; z++) {
            if (inSphere(x,y,z)) {
                cells.push_back(Cell(std::tuple<int, int, int>(xIndex() + x, yIndex() + y, zIndex() + z)));
            }
        }
    }
}
```
So we should add a unit test to make sure it gives the same result as the naive version.

But basically, I think we need to modify `ceilDistance` to take 2 arguments, so it becomes something like:

```c++
auto ceilDistance = [](int radius, int x, int y) {
    return static_cast<int>(std::ceil(std::sqrt(std::pow(double(radius), 2) - std::pow(double(x), 2) - std::pow(double(y), 2))));
};
```

And then gets used like:
```c++
int y_radius = ceilDistance(radius, x, 0);
```
```c++
int z_radius = ceilDistance(radius, x, y);
```
This way z gets correctly limited.

Jaeyoung-Lim: @jkflying Actually, I realized that I messed up. I made a simple fix by switching `radius` with `y_radius` Would this be enough? or should I try and do the cube filtering you suggested?

jkflying: This substitutes in `z_radius = sqrt(ceil(sqrt(r^2 - x^2))^2 - y^2)`, if we can ignore the `ceil` then it is `sqrt(r^2 - x^2 - y^2)` which is correct. The `ceil` makes it over-estimate a little, since we `ceil` before the sqrt/sqr cancel out. So it could be wrong by sqrt(y_radius).

If we want to be more accurate, we could do `int z_radius = ceilDistance(hypot(x_radius, y_radius), y);` and change the lambda to accept doubles.

Jaeyoung-Lim:
@jkflying Awesome, thanks!

PR #386, opened 2019-05-29
Safe Landing Planner

mrivi:
This PR adds a third planner: the Safe Landing Planner.
It runs independently from the other two and helps the vehicle landing safely.
If the terrain underneath the vehicle is flat, the vehicle will land. Otherwise, it will fly a square spiral to find a safe place to land.
It works in both Mission mode with a Land waypoint or in Auto Land.

VIDEO https://www.youtube.com/watch?v=o7nxSWULfMI


Jaeyoung-Lim:
Super Cool! Are there any videos / visualizations of it working?

mrivi:
Yes, I'll post it as soon as possible

safe_landing_planner/test/test_safe_landing_planner.cpp
```diff
@@ -0,0 +1,402 @@
...
+
+  Eigen::Vector3f pos(5.f, 5.f, 5.f);
+  safe_landing_planner.setPose(pos, q);
+
+  for (int i = 0; i < 1000; ++i) {
+    float min = 0.f;
+    float max = 9.f;
+    float x = min +
+              static_cast<float>(rand()) /
+                  (static_cast<float>(RAND_MAX / (max - min)));
```
jkflying: It would be better to use a `std::uniform_distribution`

README.md
```diff
@@ -328,6 +343,17 @@ Parameters to set through QGC:
 * Other Required Components for Intel Realsense:
   - Librealsense (Realsense SDK). The installation instructions can be found [here](https://github.com/IntelRealSense/librealsense/blob/master/doc/installation.md)
   - [Librealsense ROS wrappers](https://github.com/intel-ros/realsense.git)
+* Other Required Components for Occipital Structure Core:
+  - Download the [Structure SDK](https://structure.io/developers). Create the `build` directory and build the SDK
```
jkflying: It would be good to specify the version of the SDK it was tested with

README.md
```diff
@@ -366,6 +394,24 @@ If you would like to read debug statements on the console, please change `custom
...
 ```
 
+## Safe Landing Planner
+
+Once the catkin workspace has been built, to run the planner with a Realsense D435 and Occipital Structure Core, you can generate the launch file using the script *generate_launchfile.sh*. The script works the same as described in the section above for the Local Planner. For example:
+
+```bash
+export CAMERA_CONFIGS="camera_main,struct_core,819612070807,0.3,0.32,-0.11,0,0,0"
+export VEHICLE_CONFIG_SLP=~/catkin_ws/src/avoidance/safe_landing_planner/cfg/slpn_structure_core.yaml
+export VEHICLE_CONFIG_WPG=~/catkin_ws/src/avoidance/safe_landing_planner/cfg/wpgn_structure_core.yaml
```
jkflying: Does the waypoint generator need to be aware of the camera type? Shouldn't that be handled by the planner part?

safe_landing_planner/include/safe_landing_planner/grid.hpp
```diff
@@ -0,0 +1,79 @@
...
+#include <Eigen/Dense>
+
+namespace avoidance {
+
+class Grid {
+public:
+  Grid(const float grid_size, const float cell_size) :
+    grid_size_(grid_size),
+    cell_size_(cell_size),
+    mean_(static_cast<int>(std::ceil(grid_size/cell_size)), static_cast<int>(std::ceil(grid_size/cell_size))),
```
jkflying: I think make a member variable which captures the result of `static_cast<int>(std::ceil(grid_size/cell_size))`, then just use that. Something like `grid_cells_`

safe_landing_planner/include/safe_landing_planner/grid.hpp
```diff
@@ -0,0 +1,79 @@
...
+  ~Grid() = default;
+
+  void reset() {
+    mean_.fill(0.f);
+    variance_.fill(0.f);
+    counter_.fill(0);
+    land_.fill(0);
+  }
+
+  void resize() {
```
jkflying: This should take the new size as an argument, rather than letting the sizes be publicly available. Otherwise the size variables might be out of sync with the grids.

safe_landing_planner/include/safe_landing_planner/grid.hpp
```diff
@@ -0,0 +1,79 @@
...
+    mean_.resize(static_cast<int>(std::ceil(grid_size_/cell_size_)), static_cast<int>(std::ceil(grid_size_/cell_size_)));
+    variance_.resize(static_cast<int>(std::ceil(grid_size_/cell_size_)), static_cast<int>(std::ceil(grid_size_/cell_size_)));
```
jkflying: Can use the `grid_cells_` variable mentioned above

safe_landing_planner/include/safe_landing_planner/grid.hpp
```diff
@@ -0,0 +1,79 @@
...
+    counter_.resize(static_cast<int>(std::ceil(grid_size_/cell_size_)), static_cast<int>(std::ceil(grid_size_/cell_size_)));
+    land_.resize(static_cast<int>(std::ceil(grid_size_/cell_size_)), static_cast<int>(std::ceil(grid_size_/cell_size_)));
+    reset();
+  }
+
+  void setMean(Eigen::Vector2i &idx, float value) {
```
jkflying:
```suggestion
  void setMean(const Eigen::Vector2i &idx, float value) {
```

And similarly for 
`setVariance`
`increaseCounter`
`getMean`
`getVariance`
`getCounter`
`setFilterLimits`

safe_landing_planner/include/safe_landing_planner/safe_landing_planner.hpp
```diff
@@ -0,0 +1,112 @@
...
+
+  Eigen::Vector2i getPositionIndex() const {return pos_index_; };
+  Grid getPreviousGrid() const {return previous_grid_; };
+  Grid getGrid() const {return grid_; };
+  int getSmoothingSize() const {return smoothing_size_;};
+
+protected:
+
+  Eigen::Vector3f position_ = Eigen::Vector3f::Zero();
+  Eigen::Vector2i pos_index_ = Eigen::Vector2i(NAN, NAN);
```
jkflying: Can't initialized an int with `NAN`. Maybe use `-1` instead?

safe_landing_planner/include/safe_landing_planner/safe_landing_planner_visualization.hpp
```diff
@@ -0,0 +1,77 @@
...
+  /**
+  * @brief converts from HSV to RGB color space
+  * @param[out] fR Red component, range: [0, 1]
+  * @param[out] fG Green component, range: [0, 1]
+  * @param[out] fB Blue component, range: [0, 1]
+  * @param[in] fH Hue component, range: [0, 360]
+  * @param[in] fS Hue component, range: [0, 1]
+  * @param[in] fV Hue component, range: [0, 1]
+  */
+  void HSVtoRGB(float& fR, float& fG, float& fB, float& fH, float& fS, float& fV);
```
jkflying: It would be better to pass in and return a `std::tuple<float, float, float>`, rather than tons of arguments.
It can then be used like:
```c++
std::tie(color.r, color.g, color.b) = HSVtoRGB(std::make_tuple(h, s, v));
```

safe_landing_planner/src/nodes/safe_landing_planner_node_main.cpp
```diff
@@ -0,0 +1,15 @@
+#include "safe_landing_planner/safe_landing_planner_node.hpp"
+
+int main(int argc, char **argv) {
+  using namespace avoidance;
+  ros::init(argc, argv, "safe_landing_planner_node");
+  ros::NodeHandle nh("~");
+  SafeLandingPlannerNode NodeLSD(nh);
```
jkflying: I believe some old naming got in here

safe_landing_planner/src/nodes/waypoint_generator_node.cpp
```diff
@@ -0,0 +1,511 @@
...
+
+  pos_index_.x() = static_cast<int>(msg.curr_pos_index.x);
+  pos_index_.y() = static_cast<int>(msg.curr_pos_index.y);
+
+  grid_lsd_.setFilterLimits(position_);
+  grid_received_ = true;
+}
+
+void WaypointGeneratorNode::calculateWaypoint() {
+  updateLSDState();
```
jkflying: More old-naming stuff

mrivi:
@jkflying addressed all the comments

PR #387, opened 2019-05-29
[GlobalPlanner] Move replanning to a loop from events

Jaeyoung-Lim:
This PR moves replanning of the global planner into a fixed loop(1Hz), and makes the planner to always replan, rather than only plan when the current cell is blocked / or the path is invalid.

Previously, the replanning was only done on events(when certain topics arrive) which led to collisions when a replanning was necessary, but a sufficient event to trigger the event was missing.

This is largely because the path searching is done by assuming priors, which fails when the assumption is different with the environment. There is no mechanism to guarantee safety inbetween

### DIFF
--- a/global_planner/src/nodes/global_planner_node.cpp
+++ b/global_planner/src/nodes/global_planner_node.cpp
@@ -309,8 +309,6 @@ void GlobalPlannerNode::octomapFullCallback(const octomap_msgs::Octomap& msg) {
     ROS_INFO("  Path is bad, planning a new path \n");
     if (global_planner_.goal_pos_.is_temporary_) {
       popNextGoal();  // Throw away temporary goal
-    } else {
-      planPath();  // Plan a whole new path
     }
   }
 }
--- a/global_planner/src/nodes/global_planner_node.cpp
+++ b/global_planner/src/nodes/global_planner_node.cpp
@@ -110,7 +110,6 @@ void GlobalPlannerNode::setNewGoal(const GoalCell& goal) {
   ROS_INFO("========== Set goal : %s ==========", goal.asString().c_str());
   global_planner_.setGoal(goal);
   publishGoal(goal);
-  planPath();
 }
 
 // Sets the next waypoint to be the current goal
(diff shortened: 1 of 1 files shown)


jkflying:
Would it be better to re-trigger a planning whenever an octo-tree block status changes? Just an idea

Jaeyoung-Lim:
@jkflying That makes a lot of sense if the vehicle is able to follow the path exactly. However, currently that is not the case, so we need to be able to react to the different vehicle position, which may lead to a collision on trying to follow a (thought to be) collision path.

IMO this problem emerges because the algorithm is using a path-seraching algorithm as a path-searching planner and also a obstacle avoidance planner.

The Path searching algorithm implemented in global_planner is making unsafe assumptions of unobserved regions. Usually state of the art planners use path searching algorithm with a local avoidance algorithm that is able to evaluate collisions of short term paths. This increases the chances of finding a solution, while still ensuring safety locally. (This is still fundamentally unsafe, but is managable in low speeds)

So, for the path searching, I agree that we only need to update the path when a new update on the map is available. However, as we are also using this path for local avoidance, we need to update the path for frequently for a safe operation.

jkflying:
> The Path searching algorithm implemented in global_planner is making unsafe assumptions of unobserved regions. 

I think this is the real issue here. The only way I can think to get around it properly is to also use some kind of observation metric for the attitude so that we only fly into observed areas, and make sure areas we want to know about will be explicitly observed as soon as possible. Beyond the scope of the PR, of course, but something to think about.

global_planner/src/nodes/global_planner_node.cpp
```diff
@@ -390,9 +381,7 @@ void GlobalPlannerNode::plannerLoopCallback(const ros::TimerEvent& event) {
   }
 
   // If the current cell is blocked, try finding a path again
```
jkflying: This comment doesn't make sense any more

Jaeyoung-Lim:
@jkflying Exactly. Also limiting the path searching to observed regions will make the problem become intractable.

In my opinion, the safest way is to introduce a local planner that makes no assumptions ensures collision avoidance with faster update rates. We can then play around with the path searching algorithm while not compromising the safety of the planner.

global_planner/src/nodes/global_planner_node.cpp
Jaeyoung-Lim: I have deleted the comment

Issue #391, opened 2019-05-31
Local planner environmet vairable is not set


lioneleeedadad:
omar@omar-Lenovo-ideapad-320-15IKB:~/catkin_ws$ roslaunch local_planner local_planner_sitl_3cam.launch
... logging to /home/omar/.ros/log/5444ea26-8383-11e9-9e1b-e82a44e20f6f/roslaunch-omar-Lenovo-ideapad-320-15IKB-4026.log
Checking log directory for disk usage. This may take awhile.
Press Ctrl-C to interrupt
Done checking log file disk usage. Usage is <1GB.

environment variable 'GAZEBO_MODEL_PATH' is not set
The traceback for the exception was written to the log file

nicovanduijn:
duplicate of #383

Issue #383, opened 2019-05-29
Launch Error: environment variable 'GAZEBO_RESOURCE_PATH' is not set


Jaeyoung-Lim:
**Describe the bug**
After #381, The following error appears when launching a launchfile
```
~/catkin_ws/src/avoidance$ roslaunch global_planner global_planner_stereo.launch
... logging to /home/jaeyoung/.ros/log/298a61c6-8202-11e9-bcc9-002315d18521/roslaunch-jaeyoung-ThinkPad-T480-30070.log
Checking log directory for disk usage. This may take awhile.
Press Ctrl-C to interrupt
Done checking log file disk usage. Usage is <1GB.

environment variable 'GAZEBO_RESOURCE_PATH' is not set
The traceback for the exception was written to the log file
```

**To Reproduce**
1. Delete `GAZEBO_RESOURCE_PATH` or `GAZEBO_MODEL_PATH` from environment
2. Launch `global_planner_stereo.launch`

**Expected behavior**
Models should be able to be imported from the correct path. Reverting #381 fixes the problem.

@nicovanduijn When you said you had a problem, have you tried with deleting the `GAZEBO_MODEL_PATH` from your environmnet?  It seems that the variables need to be defined in order to make it work

nicovanduijn:
I have, yes. It worked for me, which was the reason for #381 

Did it work for you prior to #381 and then no longer?

Jaeyoung-Lim:
Yes, It worked for me prior to #381 and it is not working anymore. Reverting #381 works for me.

Wierd!

mrivi:
In the PX4 #issue channel they had the same issue 
```bash
environment variable 'GAZEBO_RESOURCE_PATH' is not set
```

Can we revert this one? Think more about this change... Setting the `GAZEBO_MODEL_PATH` and `GAZEBO_RESOURCE_PATH` doesn't ensure the planners running correctly since the ` ~/Firmware/Tools/setup_gazebo.bash` also sets them.

nicovanduijn:
I really don't understand what's going on. It seems that what works on my machine is orthogonal to what works on Jay's...

nicovanduijn:
It looks like my machine is the odd one, and pre- #381 works for most. I'll revert it and figure out what's wrong on my machine

mrivi:
fixed by https://github.com/PX4/avoidance/pull/392

PR #378, opened 2019-05-27
Add gazebo_model_path to launchfile

Jaeyoung-Lim:
This PR adds the environment variable `GAZEBO_MODEL_PATH` and `GAZEBO_RESOURCE_PATH` to the launchfile, which eliminates the need to specify the environment as 
```
export GAZEBO_MODEL_PATH=${GAZEBO_MODEL_PATH}:~/catkin_ws/src/avoidance/avoidance/sim/models
```

For the lazy people

PR #392, opened 2019-05-31
Yet another GAZEBO_MODEL_PATH fix

nicovanduijn:
Ok, let's figure this out now. 

I think setting GAZEBO_MODEL_PATH in the bashrc is the most portable thing and will work on everyone's machine.

Can you please test @Jaeyoung-Lim, so I don't break your setup once again? (Sorry...)

Also, let me know if the Readme is not clear enough about the installation

Addresses #383 #391

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,6 +172,10 @@ In the following section we guide you trough installing and running a Gazebo sim
    ```bash
    export ROS_PACKAGE_PATH=${ROS_PACKAGE_PATH}:~/Firmware
    ```
+1. Finally, set the GAZEBO_MODEL_PATH in your bashrc:
+```bash
+echo export GAZEBO_MODEL_PATH=${GAZEBO_MODEL_PATH}:~/catkin_ws/src/avoidance/avoidance/sim/models:~/catkin_ws/src/avoidance/avoidance/sim/worlds >> ~/.bashrc
+```
 
 You should now be ready to run the simulation using local or global planner.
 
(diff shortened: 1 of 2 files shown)


lioneleeedadad:
how can I set GAZEBO_MODEL_PATH in the bashrc ?

nicovanduijn:
@lioneleeedadad did you look at the actual PR? It includes a Readme update

jkflying:
On my system before I didn't even have a GAZEBO_RESOURCE_PATH, so I think just without that in the launch file it should work fine. The reason we got errors is because it tried to append to a variable that didn't exist yet, and launch files are more fussy than bash.

I'd prefer for it to stay in the launch file, both to simplify setup and also to not have stuff in people's .bashrc just because they ran the planner once.

baumanta:
I had to remove the two lines from the launch file as well on my setup, otherwise I get an error that the PATH is not set

nicovanduijn:
Does this work for you @Jaeyoung-Lim ?

Jaeyoung-Lim:
@nicovanduijn This works for me. If I understand correctly, this is reverting #378 #381 ?

nicovanduijn:
Essentially, yes

Jaeyoung-Lim:
Can we get this merged anyway? Current master is just broken for me...Is this also the case for others? or is this just me?

nicovanduijn:
lgtm

PR #394, opened 2019-05-31
use correctly stamped tf for pointcloud

baumanta:
This makes the planner use the correct TF to transform the pointcloud (at the same time a s the PC stamp). Before it just used the latest TF.

This is possible now, as we listen for the tf in a separate thread. Before that, when we attempted this, we used to get a lot of timeouts on the planner as the TF was blocking. Now it is possible to use the correct TF without getting any timeouts.


baumanta:
I'm still a bit sceptical about this. In one of 3 test runs we saw timeouts. Needs more testing

local_planner/src/nodes/local_planner_node.cpp
```diff
@@ -179,10 +178,8 @@ size_t LocalPlannerNode::numReceivedClouds() {
 size_t LocalPlannerNode::numTransformedClouds() {
   size_t num_transformed_clouds = 0;
   for (size_t i = 0; i < cameras_.size(); i++) {
-    std::unique_lock<std::mutex> lk(*(cameras_[i].trans_ready_mutex_));
-    cameras_[i].trans_ready_cv_->wait_for(
-        lk, std::chrono::milliseconds(30),
-        [this, i] { return cameras_[i].transformed_; });
+    std::lock_guard<std::mutex> transformed_cloud_guard(
+        *(cameras_[i].transformed_cloud_mutex_));
```
sunzj: Since localplanner_node running period is 100ms. https://github.com/PX4/avoidance/blob/master/local_planner/src/nodes/local_planner_node.cpp#L21

Just checking transform status(don't wait for transform finish) may cause don't have chance to run planner. Then the interval of planning will be longer than 100ms.

local_planner/src/nodes/local_planner_node.cpp
```diff
@@ -532,14 +518,10 @@ void LocalPlannerNode::printPointInfo(double x, double y, double z) {
 
 void LocalPlannerNode::pointCloudCallback(
     const sensor_msgs::PointCloud2::ConstPtr& msg, int index) {
+  std::lock_guard<std::mutex> lck(*(cameras_[index].cloud_msg_mutex_));
   cameras_[index].newest_cloud_msg_ = *msg;  // FIXME: avoid a copy
   cameras_[index].received_ = true;
-
-  {
-    std::unique_lock<std::mutex> lck(*(cameras_[index].cloud_ready_mutex_));
-    cameras_[index].transformed_ = false;
```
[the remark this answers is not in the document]
baumanta: I think it should work never the less: You are right, if a new cloud is comming in before the last transformed one is copied into the planner `transformed_` is still true. This means that the cloud can still be copied into the planner (which is ok, because `cameras_[index].pcl_cloud` still contains the last transformed cloud) the transforming of the new cloud will be blocked until the last one is copied into the planner and `transformed_` is set to false.

Do you agree?

local_planner/src/nodes/local_planner_node.cpp
baumanta: this function is called from the main thread which also does all the queries on the rostopics (e.g. getting new pointclouds) if we block that until it is transformed we will not get any new pointclouds anymore. So we would try to transform a specific pointcloud until done, instead of trying to transform the most recent one if it takes too long.
Blocking at that point could even lead to a deadlock if the transformation of a specific pointcloud is impossible (dropped data, ...).

local_planner/src/nodes/local_planner_node.cpp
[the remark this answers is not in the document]
sunzj: Yes, so if the new frame comes, it still use last transformed frame.

local_planner/src/nodes/local_planner_node.cpp
sunzj: The new pointcouds is triggered by callback, won't block by this function.
And it won't causes deadlock, since there is a max wait threshold time.

baumanta: The callback for the pointcloud is queried in the same thread, so if we wait at that location we will not get new pointclouds during the waiting period.
You are right, the max time works for one camera, but if we have multiple, we would also block for potentially x-times that timespan. Therefore, I would rather leave it that way - spinning around once more if the pointcloud is not yet transformed.

sunzj: I still not understand why it will causes cant get new cloudpoints issue. Waiting time out happen for multiple cameras,that's impossible,since the camera is 30fps.

And spinning once more will increase 100ms for interval of running planner.

baumanta: To why we do not get clouds in this time interval:

every 100ms `cmdLoopCallback` is executed. At the beginning of that function all the callbacks are evaluated: https://github.com/PX4/avoidance/blob/aeed406574119fb43a09958c82d28e0f52fc287b/local_planner/src/nodes/local_planner_node.cpp#L290

Next comes `updatePlanner()` which contains the waiting. Only when the waiting is finished and `cmdLoopCallback` is executed again, the new pointclouds are received

baumanta: We could technically wait those 30ms, that would be fine, but as soon as we use 3 cameras we would be at a max timeout of 90ms, as soon as that time is close to 100ms or higher we will slow down the main loop

sunzj: ...the max timeout 90ms may happen, however, the possibility is very small... since there are 3 threads to handle each camera independently.
The callback trigger at the beginning of `cmdLoopCallback` may cause the latest cloud points not use in planner. Since cmdLoopCallback to updatePlanner() is very fast. Callback send the new point cloud then transform the point cloud need more time. The lastest point cloud may not be used in updatePlanner().
Anyway, i didn't test it yet....

jkflying: But since the camera frame rate is higher than the planner, and the transform is very fast, this way the planner always has the latest transformed point cloud available. It just gets whatever is there, even if maybe another one will be available soon. Otherwise the planner is delayed.

PR #390, opened 2019-05-31
[GlobalPlanner] Deprecate launchfiles and update readme

Jaeyoung-Lim:
This PR deprecates launchfiles that are unused / obsolete in the global_planner, and update the README accordingly

PR #395, opened 2019-05-31
Global planner launchfile unification / depth camera model

Jaeyoung-Lim:
This PR enables a depth camera model in global_planner with a launchfile `global_planner_depth-camera.launch` as discussed in #390 @mrivi 

In order to share as much code as possible between local and global_planner, a `avoidance_sitl_stereo.launch` file was added that is both used by local and global planner


avoidance/launch/avoidance_sitl_stereo.launch
```diff
@@ -0,0 +1,28 @@
+<launch>
+  <arg name="model" default="iris_depth_camera"/>
```
mrivi:
```suggestion
  <arg name="model" default="iris_stereo_camera"/>
```

PR #396, opened 2019-06-03
Make depth image at 30Hz then downsample

nicovanduijn:
This commit changes the auto-generation of the launch file, as well
as the realsense launch file to generate a depth image at 30Hz, then
downsample it to 15Hz before sending to the nodelet. Doing so
fixes a realsense-related bug that breaks auto-exposure on USB2.0


nicovanduijn:
Closing, this doesn't help

PR #399, opened 2019-06-06
Stable

bzt:
navigation/global_planner clashes with avoidance/global_planner
avoidance/global_planner needs to be renamed since navigation/global_planner is older


Jaeyoung-Lim:
@bzt Could you clarify on what the package `navigation` you are referring to? Are you referring to the [ros-planning](https://github.com/ros-planning/navigation)?

bzt:
http://wiki.ros.org/navigation

https://github.com/ros-planning/navigation

bzt:
our ros stack serves flying and driving drones from one workspace. navigation/global-planner may be not the best option for driving drones, but it is a good start to work on for our apprentices, working students or new colleagues. 

so we need navigation/global-planner to be able to build in one workspace side-by-side with avoidance/avoidance-global-planner.

jkflying:
If we are going to rename, a better name would be `px4_global_planner`, also then to be consistent we should also rename the `local_planner` to `px4_local_planner`. This will have a lot of downstream effects too.

Only the node name needs to change, as well. The launch files, directories etc can all stay the same.

bzt:
sounds right, (sorry for the launch files, but a complete directory search n replace was a quick fix on our side)

and.. but. just renaming the nodes wont work since you guys use some clever .cfg to .h stuff unknown to us.

just renaming the package in the package xml and cmakelist does not work with a lot of work involved. (because the cfg to h stuff then searches the wrong include path and fails at building).

we were not able to accomplish that and just rename everything, which we think is a save fix ;) (launch file names should stay the same, agreed;)

TobiMiller:
+1 for renaming.
@bzt you need both planner on your robot at the same time? seperate workspaces without sourcing each of them is not an option?

bzt:
would be possible, but why you were naming your package as an already existing package in the first place ?

separate workspaces are possible but against having one workspace that suits all needs, either flying or driving drones.

i think better stick to never name your package the same as an existing one.. 

so for the future live of px4 avoidance just rename it now, which will save a lot of work in time comijg, dont you think so ?

bzt:
i believe you guys could occupy local_planner as an package..(we did not find a package with that name yet;)

which i intend to believe is anyway the better option on avoidance for flying drones.

Jaeyoung-Lim:
I agree that the namings are not ideal as @TobiMiller has pointed out. 

However, I don't see why we need to resolve a naming conflict with the `ros/navigation` stack that has no usecase. As @bzt has confirmed that this is not coming from a use case but simply a preference to put both packages in the same workspace. 

In my opinion, there are no use cases that you need both packages in the same workspace. The PX4/avoidance is not depending on any of the ros/navigation components. Therefore, the chaos that the naming change will bring is not really worth the benefits we get.

bzt:
which only tells you guys are arrogant and think you could override naming conventions in the ros world..

good work. keep on lazy doin it..

bzt:
we are the use case for both worlds in one workspace.

Jaeyoung-Lim:
@bzt I apologize if it sounded arrogant, as I was the only person opposed to the idea I assume 'you guys' mean myself.

The ros package naming [conventions](http://wiki.ros.org/ROS/Patterns/Conventions) in the ros world mention that the prefixes should be only used when the package is used widely. This is the reason why I asked you to clarify why it is necessary to have the `px4/avoidance` package and the `ros/navigation` package in the same workspace, as `px4/avoidance` and `ros/navigation` have a similar functionality and it is hard for me to imagine why you would want packages with redundent functionalities in the same workspace. `px4/avoidance` is a end application, and cannot be used as a component of a larger system. (at least for now)

If there is a proper use case worth fixing, we certainly need to invest time to fix this.

bzt:
again, it was a mistake to name a package as an already existing one. the use case for navigation/global_planner is always given since it is one of the most basic packages in the ros world, i believe everyone who starts with ros comes to this package on a certain time.

and you now suggest a workspace build from ground up has to be destroyed as soon as someone wants to use the avoidance/global_planner in that workspace just because someone was to lazy to do his job and research that his package naming does not clash, really ??!

jkflying:
@bzt please keep this civil.

Jae has a point. Adding the PX4 global_planner to your workspace is not sufficient to do planning. This planner is tightly integrated with the PX4 / MAVROS system, and can't just be substituted for a different planner from a generic ROS package.

I understand that you want to add it to a single workspace with all the planners you can find, but without all of the other integration work there really is no point - you can't just take a different launch file and then control a drone, you also need MAVROS, SITL/a real fcu, etc. You will need these components sourced into your ROS environment, and they have a lot of additional requirements.

Basically, I think you need to really evaluate if having these different planners in the same environment actually makes sense. So far from what I see it doesn't make sense to have them together in the same environment, there is no use case where you would ever use them together, and they don't even support the same types of sensors or vehicles.

bzt:
please unsubscribe me from this conversation, discussion makes no sense, it is not difficult to google ‚ros global_planner‘ before deploying a release.

we will delete your .git folder and stick to our changes, before we destroy a workspace that is using the navigation stack on almost a daily schedule.

TobiMiller:
@jkflying @Jaeyoung-Lim 
dont feed the troll. He doesn't understand that it's open source here and nobody is obligated. I don't think he's good at ros yet and lets his frustration out on others.

peace out.

PR #400, opened 2019-06-07
Launchfile required

baumanta:
put required tags in auto generated launch file, such that if one critial node dies, all of them dies immediatly. This assists keeping everything running together.

Dependent on https://github.com/IntelRealSense/realsense-ros/pull/804


baumanta:
We have to move on, so we'll merge this even though the PR on the Realsense repo did not yet get in. If you get an error on `argument required does not exist or similar` when launching an autogenerated launchfile, pleas cherry-pick the mentoned commit onto your realsense wrapper

PR #402, opened 2019-06-10
Export liblocal_planner.so for other projects

jschwarz89:
This patch introduces a minor change to local_planner's CMakeLists.txt,
which allows other cmake projects to use the library as an external
project. Without this change, compiling against local_planner introduces
the following errors (to begin with):

undefined reference to `avoidance::LocalPlanner::LocalPlanner()'
undefined reference to `avoidance::LocalPlanner::~LocalPlanner()'

### DIFF
--- a/local_planner/CMakeLists.txt
+++ b/local_planner/CMakeLists.txt
@@ -106,6 +106,7 @@ generate_dynamic_reconfigure_options(
 ## DEPENDS: system dependencies of this project that dependent projects also need
 catkin_package(
   INCLUDE_DIRS include
+  LIBRARIES local_planner
   CATKIN_DEPENDS roscpp rospy std_msgs mavros_msgs geometry_msgs sensor_msgs message_runtime tf avoidance
 #  DEPENDS system_lib
 )


jkflying:
Hi, thanks for the PR. It seems CI is failing, that isn't due to this PR but we'd like to have it working before merging.

Long term I'd like to switch to a static library, would that be an issue for you?

jschwarz89:
> Hi, thanks for the PR. It seems CI is failing, that isn't due to this PR but we'd like to have it working before merging.
> 
Sounds reasonable - can we maybe re-trigger the CI if it's not related to my issue? I'm not familiar with your CI at all...

> Long term I'd like to switch to a static library, would that be an issue for you?
Nope - as long as I can link against it, I'm a happy droner :-)

jkflying:
I've re-triggered a few times, turns out it is due to https://discourse.ros.org/t/new-gpg-keys-deployed-for-packages-ros-org/9454, I will get that fixed ASAP.

I see from Slack you are interested in the point cloud -> laser scan code, this will probably be factored out into a generic `avoidance` package, so at some point in the hopefully-not-to-distant future you will need to link against that instead.

jkflying:
Ok, key issue is fixed, if you could rebase this we can run CI and hopefully get it merged :+1:

jschwarz89:
> I see from Slack you are interested in the point cloud -> laser scan code, this will probably be factored out into a generic `avoidance` package, so at some point in the hopefully-not-to-distant future you will need to link against that instead.
>

That sounds awesome, I'll stay tuned. Anyways, rebased with master :-)

PR #404, opened 2019-06-11
Update ROS key due to upstream key change

jkflying:
https://discourse.ros.org/t/new-gpg-keys-deployed-for-packages-ros-org/9454

This is a blocker for https://github.com/PX4/avoidance/pull/402


nicovanduijn:
I guess this same change should be reflected in the readme?

jkflying:
Yep, missed that.

PR #406, opened 2019-06-12
Limit speed to what is safe given box size, accel, jerk

jkflying:
This way you can't configure maximum speed to something that the planner won't be able to stop in time for (at least theoretically :wink: ).

### DIFF
--- a/local_planner/src/nodes/local_planner.cpp
+++ b/local_planner/src/nodes/local_planner.cpp
@@ -305,9 +305,19 @@ void LocalPlanner::getObstacleDistanceData(
 avoidanceOutput LocalPlanner::getAvoidanceOutput() const {
   avoidanceOutput out;
   out.waypoint_type = waypoint_type_;
-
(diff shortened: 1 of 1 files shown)


local_planner/src/nodes/local_planner.cpp
```diff
@@ -305,9 +305,19 @@ void LocalPlanner::getObstacleDistanceData(
 avoidanceOutput LocalPlanner::getAvoidanceOutput() const {
   avoidanceOutput out;
   out.waypoint_type = waypoint_type_;
-
   out.obstacle_ahead = !polar_histogram_.isEmpty();
-  out.cruise_velocity = px4_.param_mpc_xy_cruise;
+
+  // calculate maximum speed given the sensor range and vehicle parameters
+  // quadratic solve of 0 = u^2 + 2as, with s = u * |a/j| + r
```
mrivi: maybe add in the comment what the letters in the formula stand for?

PR #411, opened 2019-06-17
[GlobalPlanner] Lower replanning rate

Jaeyoung-Lim:
This is a temporary fix for the segmentation faults that were being observed in the global_planner.

Lowering the planner loop seems to solve the problem for now


Jaeyoung-Lim:
@jkflying Can we get this merged first while I investigate on the race condition?

PR #413, opened 2019-06-19
Add libsensors_monitor for launchfile generation

Jaeyoung-Lim:
This PR adds the libsensors_monitors package to the generated launchfile. 

[libsensors_monitor](http://wiki.ros.org/libsensors_monitor) is a package used to monitor temperature information on the companion computer.

### DIFF
--- a/safe_landing_planner/tools/generate_launchfile.sh
+++ b/safe_landing_planner/tools/generate_launchfile.sh
@@ -109,6 +109,11 @@ EOM
 echo "Adding vehicle paramters: $VEHICLE_CONFIG_WPG"
 fi
 
+cat >> launch/safe_landing_planner_launch.launch <<- EOM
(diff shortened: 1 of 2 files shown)


mrivi:
@Jaeyoung-Lim could you please add it also here https://github.com/PX4/avoidance/blob/master/safe_landing_planner/tools/generate_launchfile.sh

Jaeyoung-Lim:
Done!

mhkabir:
Nice addition, but why should be live in tree with avoidance?

Jaeyoung-Lim:
@mhkabir the launchfile naming is incorrect. In only launches the `local_planner`

dayjaby:
Can we keep the dependencies minimal? If you need those libsensors just put it in a private launch file.

jkflying:
Agreed, this should go in a separate launch file.

PR #420, opened 2019-06-26
Compute FOV in camera frame

nicovanduijn:
This commit moves the dynamic calculation of the FOV into the camera
frame, such that it is not affected by faulty transforms.

This removes the need to subscribe to the camera_info topic.

Furthermore, we no longer make the assumption that the FOV consists of a central forward-pointing camera, but we can account for multiple cameras, with overlapping FOV or gaps.

image

The adaptSpeed() function now scales the waypoint only toward the "true" edge of the FOV, meaning that it will not be scaled when moving from one camera to the other, as long as they have adjacent or overlapping FOV

### DIFF
--- a/avoidance/include/avoidance/common.h
+++ b/avoidance/include/avoidance/common.h
@@ -1,6 +1,8 @@
 #ifndef COMMON_H
 #define COMMON_H
 
+#include "avoidance/histogram.h"  // needed for ALPHA_RES
+
 #include <pcl/point_cloud.h>
 #include <pcl/point_types.h>
 #include <Eigen/Core>
@@ -77,7 +79,7 @@ bool pointInsideFOV(const FOV& fov, const PolarPoint& p_pol);
 /**
 * @brief      compute in which FOV the current point lies
 * @param[in]  vector of FOV defining the field of view of the drone
-* @param[in]  polar point of the current orienation in question
+* @param[in]  polar point of the current orientation in question
 * @param[out] index pointing to the camera in the FOV struct which contains the
 *             current point
 * @returns    boolean value if the point in question is in exactly one FOV
@@ -108,8 +110,6 @@ bool isOnEdgeOfFOV(const std::vector<FOV>& fov_vec, const PolarPoint& p_pol, int
 *            go
 * @returns   a scale [0, 1] depending on whether the point in question can be
 *            seen from here
-* @TODO:     currently this is binary depending on whether its inside or outside
-*            in the near future I want to scale this correctly
 **/
 float scaleToFOV(const std::vector<FOV>& fov, const PolarPoint& p_pol);
 
--- a/avoidance/include/avoidance/common.h
+++ b/avoidance/include/avoidance/common.h
@@ -83,6 +83,34 @@ bool pointInsideFOV(const std::vector<FOV>& fov_vec, const PolarPoint& p_pol);
 bool pointInsideFOV(const FOV& fov, const PolarPoint& p_pol);
 
 /**
+* @brief      compute in which FOV the current point lies
+* @param[in]  vector of FOV defining the field of view of the drone
+* @param[in]  polar point of the current orienation in question
+* @param[out] index pointing to the camera in the FOV struct which contains the
+*             current point
+* @returns    boolean value if the point in question is in exactly one FOV
+* @warning    This function returns false and sets the index to -1 if there is
+*             no or more than one camera which sees the current point
+**/
+bool isInWhichFOV(const std::vector<FOV>& fov_vec, const PolarPoint& p_pol,
+                  int& idx);
+
+/**
+* @brief      determine whether the given point lies on the edge of the field
+*             of view or between two adjacent cameras
+* @param[in]  vector of FOV defining the field of view of the drone
+* @param[in]  polar point of the current orientation in question
+* @param[out] index of the camera in the FOV vector, indicating which FOV edge
+*             it is on, if any. -1 if none
+* @returns    boolean indicating whether the current point is on the edge of the
+*             field of view
+* @warning    This function returns false and sets the index to -1 if the point
+*             is not on the edge of the fov
+**/
+bool isOnEdgeOfFOV(const std::vector<FOV>& fov_vec, const PolarPoint& p_pol,
+                   int& idx);
+
+/**
 * @brief     function returning a scale value depending on where a polar point
 *            is relative to the field of view
 * @param[in] vector of FOV structs defining the field of view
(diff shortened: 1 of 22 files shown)


avoidance/src/common.cpp
```diff
@@ -338,4 +338,38 @@ void fillUnusedTrajectoryPoint(mavros_msgs::PositionTarget& point) {
...
+void removeNaNFromPointCloud(pcl::PointCloud<pcl::PointXYZ>& cloud, FOV& fov) {
+  // Filter out NANs and keep track of outermost points for FOV
+  size_t j = 0;
+  float h_max = -9999.f, v_max = -9999.f, h_min = 9999.f, v_min = 9999.f;
+  for (size_t i = 0; i < cloud.points.size(); ++i) {
+    if (!std::isfinite(cloud.points[i].x) ||
+        !std::isfinite(cloud.points[i].y) || !std::isfinite(cloud.points[i].z))
+      continue;
+    cloud.points[j] = cloud.points[i];  // safe, because i is always ahead of j
+    h_max = std::max(h_max, cloud.points[i].x / cloud.points[i].z);
```
jkflying: Divisions are really slow. I would instead suggest to check if the value is bigger than h_max via `h_max * z < x `, and then if so assign the division result.

avoidance/include/avoidance/common.h
```diff
@@ -214,6 +215,16 @@ void transformVelocityToTrajectory(mavros_msgs::Trajectory& obst_avoid,
...
-}
+
+/**
+* @brief           This is a refactored version of the PCL library function to
+*                  remove NAN values from the point cloud and compute the FOV
+* @note            It operates in-place and iterates through the cloud once
+* @param[in, out]  cloud The point cloud to be filtered in the camera frame
+* @param[in, out]  fov of the camera
+**/
+void removeNaNFromPointCloud(pcl::PointCloud<pcl::PointXYZ>& cloud, FOV& fov);
```
jkflying: This also calculates the FOV, so I'd stick that in the name somewhere too

avoidance/src/common.cpp
```diff
@@ -338,4 +338,38 @@ void fillUnusedTrajectoryPoint(mavros_msgs::PositionTarget& point) {
   point.yaw = NAN;
   point.yaw_rate = NAN;
 }
+
+// This function is a refactor of the original in the pcl library
+void removeNaNFromPointCloud(pcl::PointCloud<pcl::PointXYZ>& cloud, FOV& fov) {
```
jkflying: Unit test the FOV logic of course

avoidance/src/common.cpp
nicovanduijn: great idea! I put a good bit of thought into how to make this most efficient, but I hadn't thought of that. Main idea behind pulling this out of PCL was to avoid yet another loop through the entire point cloud

avoidance/include/avoidance/common.h
nicovanduijn: agreed

avoidance/src/common.cpp
nicovanduijn: done, check my latest commit

avoidance/src/common.cpp
```diff
@@ -8,24 +8,90 @@
...
+}
+
+bool isOnEdgeOfFOV(const std::vector<FOV>& fov_vec, const PolarPoint& p_pol, int& idx) {
+  idx = -1;
+  bool retval = false;
+  if (isInWhichFOV(fov_vec, p_pol, idx)) {
+    PolarPoint just_outside = p_pol;
+    // todo: check for pitch!
+    if (wrapAngleToPlusMinus180(p_pol.z - fov_vec[idx].yaw_deg) > 0.0f) {  // to the right
+      just_outside.z = wrapAngleToPlusMinus180(fov_vec[idx].yaw_deg + fov_vec[idx].h_fov_deg / 2.0f + 1.0f);
```
baumanta: could you explain the number 1.0f? Why does being 1 deg outside have any significant meaning in contrast to being bigger than the exact edge or for example 5 degrees?

avoidance/src/common.cpp
```diff
@@ -315,4 +393,105 @@ void fillUnusedTrajectoryPoint(mavros_msgs::PositionTarget& point) {
   point.yaw = NAN;
   point.yaw_rate = NAN;
 }
+
+// This function is a refactor of the original in the pcl library
+pcl::PointCloud<pcl::PointXYZ> removeNaNAndGetMaxima(pcl::PointCloud<pcl::PointXYZ>& cloud) {
```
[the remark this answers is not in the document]
nicovanduijn: Depends on whether the global planner would also be interested in receiving NAN-filtered point clouds?

avoidance/src/common.cpp
nicovanduijn: Yeah, that's an arbitrary magic number, which is nasty. I couldn't think of a better way to do this. (Can you?)
I'll add a comment in the meantime and spend some time thinking of a better way to do this

avoidance/src/common.cpp
```diff
@@ -315,4 +393,105 @@ void fillUnusedTrajectoryPoint(mavros_msgs::PositionTarget& point) {
...
+    v_max = std::max(p_pol_fcu.e, v_max);
+  }
+
+  float h_diff = std::min(h_max - h_min, 360.0f - h_max + h_min);
+  float v_diff = std::min(v_max - v_min, 360.0f - v_max + v_min);
+
+  if (h_diff > fov.h_fov_deg) {
+    fov.h_fov_deg = h_diff;
+
+    // Note: the wrapping here assumes the FOV < 180 degrees!
```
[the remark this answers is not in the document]
nicovanduijn: done

avoidance/src/common.cpp
nicovanduijn: I tried adding FLT_MIN and FLT_EPSILON, but that doesn't work, because those constants are only significant around zero and get rounded away for larger numbers due to the inherent nonuniform distribution of floating point numbers.

I could tie it to ALPHA_RES somehow, since that somewhat meaningfully represents the smallest angle increment we care about? Any ideas?

jkflying: I like the idea of a minimum FOV being ALPHA_RES + eps.

nicovanduijn: downside is I have to include histogram.h in common.h. I hate introducing these dependencies, but it's equally ugly to re-define that resolution in another place

baumanta: yes that sounds better, if you exchange the 1.0 by ALPHA_RES/2  that would kinda make sense

avoidance/src/common.cpp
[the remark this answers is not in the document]
baumanta: yes I agree that this function could be useful as well for global planner. I think it can stay where it is for now. I just think that once we have more of these bigger functions in common.cpp it might make sense to split that file again into super basic stuff and the more specialized functions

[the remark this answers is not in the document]
nicovanduijn: Yes, I thought about a FOV class with respective member functions. Not sure if it's worth it

PR #421, opened 2019-06-26
poll px4 paramters at a higher rate if they aren't set

mrivi:
Issue in Master: the first time we try to retrieve the parameters we get
``` [ERROR] [1561541271.521030400, 0.420000000]: PR: Unknown parameter to get: MPC_XY_CRUISE
[ERROR] [1561541271.523737193, 0.424000000]: PR: Unknown parameter to get: MPC_COL_PREV_D
 ```
The next poll happens 30 seconds later which is a huge time delay. 

This is more a workaround to make sure that the CI is working. It would be nice to investigate why the first attempt always fails.


jkflying:
Does this fix the time delay issue on the Firmware CI?

nicovanduijn:
how does this relate to #369 ? Wouldn't that do the trick as well?

mrivi:
Yes, CI is now working. 
The drawback of #369 is that for the first 30seconds you could be flying with out of date parameters since the default on the Firmware side change continuously and it is impossible to keep the two codebases in sync. 
I feel we should more get to the bottom of why the first attempt fails

PR #423, opened 2019-06-27
slp: use command associated with wp instead of mission item list

mrivi:


jkflying:
This is blocked by a MAVROS release.

safe_landing_planner/src/nodes/waypoint_generator_node.cpp
```diff
@@ -83,7 +82,7 @@ void WaypointGeneratorNode::trajectoryCallback(const mavros_msgs::Trajectory &ms
   if (update && msg.point_valid[0] == true) {
     waypointGenerator_.goal_ = avoidance::toEigen(msg.point_1.position);
     waypointGenerator_.velocity_setpoint_ = avoidance::toEigen(msg.point_1.velocity);
-
+    waypointGenerator_.is_land_waypoint_ = (msg.command[1] == 21);
```
jkflying: I know we do this in a bunch of places, but I'd really prefer to have a big list of what the commands are (either as `#define`s or as enums) so that we aren't putting codes like 21 or 178 all over our codebase. It makes the code easier to read, and means we have a single place to update if we find we made a mistake or want to use a new message or something.

mrivi:
Note to self: something like this https://github.com/PX4/avoidance/pull/445/commits/25d72b1ab9f416a18b35f6753cf32771447ebe32 needs to be done here otherwise we are not compatible with default line tracking

PR #426, opened 2019-07-01
clip depth camera at 18m instead of 9

mrivi:
On the local planner we require a box size of 12m but actually in the model we're clipping at 9m.

Master:
before

PR:
after

Needs testing on ROS Kinetic since the problem was showing up only in Melodic


nicovanduijn:
Your first screenshot of master looks a bit concerning... what's wrong there?

jkflying:
Works on Kinetic. The issue with the first one is that the sensor changed slightly in Melodic, so that if there isn't anything at that distance it gives the max distance instead of no points. So we were getting lots of bad points without this fix in Melodic.

PR #430, opened 2019-07-04
Create common px4 config for avoidance

Jaeyoung-Lim:
Previously, the px4 configuration in the `avoidance_sitl_mavros.launch` was located under the local planner. However, this launch file is used by both local and global planner. This required to build local_planner in the workspace to use global_planner

This PR adds a new px4_config.yml configuration file that is under the avoidance package that can be shared by all planners. The contents are currently identical with the config file in local planner. I have not deleted the config file under local planner in case we need local_planner specific configuration of the firmware


avoidance/launch/avoidance_sitl_mavros.launch
```diff
@@ -24,7 +24,7 @@
         <include file="$(find mavros)/launch/node.launch">
             <arg name="pluginlists_yaml" value="$(find mavros)/launch/px4_pluginlists.yaml" />
             <!-- Need to change the config file to get the tf topic and get local position in terms of local origin -->
-            <arg name="config_yaml" value="$(find local_planner)/resource/px4_config.yaml" />
+            <arg name="config_yaml" value="$(find avoidance)/resource/px4_config.yaml" />
```
mrivi: can you use [this one](https://github.com/mavlink/mavros/blob/master/mavros/launch/px4_config.yaml) instead?

Jaeyoung-Lim: Thanks for the feedback. Much better!

Jaeyoung-Lim:
@mrivi I have tested against local and global planner.

mrivi:
@Jaeyoung-Lim cool, thanks! Safe landing planner is working as well.
Do you mind updating the generate_launchfile.sh for both the local and safe landing as well?

Jaeyoung-Lim:
@jkflying Can we get this merged? I am repeatedly running into this problem when trying to clean up the workspace and build global_planner

PR #431, opened 2019-07-04
Port avoidance_node to local_planner

Jaeyoung-Lim:
This PR ports `avoidance_node` to `local_planner` which makes the companion process status publishing run independently on a separate queue.

The global_planner has already been using the `avoidance_node`, and it would be useful to have a single node that takes care of the companion process status communication as well as failsafe checks


mrivi:
@Jaeyoung-Lim - CI is failing due to the failsafe tests https://github.com/PX4/avoidance/blob/master/local_planner/test/test_local_planner_node.cpp

mrivi:
Cherry-picked on  #434

PR #437, opened 2019-07-08
Local Planner: Various fixes

mrivi:
Fixes:
- planner restart
- RTL
- multiple takeoffs in mission without power cycle

### DIFF
--- a/local_planner/include/local_planner/local_planner.h
+++ b/local_planner/include/local_planner/local_planner.h
@@ -54,6 +54,7 @@ class LocalPlanner {
   float max_point_age_s_ = 10;
   float yaw_fcu_frame_deg_ = 0.0f;
   float pitch_fcu_frame_deg_ = 0.0f;
+  float dist_to_goal_xy_ = 100.f;
 
   std::vector<FOV> fov_fcu_frame_;
 
--- a/local_planner/include/local_planner/local_planner.h
+++ b/local_planner/include/local_planner/local_planner.h
@@ -125,7 +125,6 @@ class LocalPlanner {
 
   ModelParameters px4_;  // PX4 Firmware paramters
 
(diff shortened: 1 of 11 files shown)


local_planner/src/nodes/local_planner.cpp
```diff
@@ -61,6 +61,9 @@ void LocalPlanner::dynamicReconfigureSetParams(avoidance::LocalPlannerNodeConfig
 void LocalPlanner::setGoal(const Eigen::Vector3f& goal) {
   goal_ = goal;
   ROS_INFO("===== Set Goal ======: [%f, %f, %f].", goal_.x(), goal_.y(), goal_.z());
+  Eigen::Vector2f pos_xy = position_.head<2>();
+  Eigen::Vector2f goal_xy = goal_.head<2>();
+  pos_to_goal_dist_xy_ = (position_ - goal_).head<2>().norm();
```
nicovanduijn: I recall defining a macro for this at some point...
Consider moving it to common so we can use it everywhere?
https://github.com/PX4/avoidance/blob/3acc9a55d7a57e0c34919087637ea50520bf045b/local_planner/src/nodes/waypoint_generator.cpp#L8

local_planner/include/local_planner/local_planner.h
```diff
@@ -52,6 +52,7 @@ class LocalPlanner {
   float min_realsense_dist_ = 0.2f;
   float smoothing_margin_degrees_ = 30.f;
   float max_point_age_s_ = 10;
+  float pos_to_goal_dist_xy_ = 100.f;
```
nicovanduijn: can you call it `dist_to_goal_xy_`? Since `pos` seems to imply a vector rather than a scalar

PR #438, opened 2019-07-09
Avoidance 0.3 alpha

jkflying:
Issues still to fix for final release:
- Planning not performed all the way to sensor horizon

Highlights:
- Safe landing planner (only compatible with post-1.9 firmware)
- Add support for Occipital Structure Core
- Update to latest Intel Realsense wrapper

Local Planner:
- Pointclouds transformed in background, allowing more rapid planning in poor comms environments
- Query parameters from flight controller to set things like cruise speed
- Time-accurate pointcloud transformations using custom-cached transforms
- Accurate history for use with narrow-FOV sensors
- Accurate camera FOV auto-estimation
- Many small bugfixes and performance improvements

Global Planner:
- Use shared components from Local Planner for heartbeats etc.

Note: this is the final release of the avoidance compatible with Firmware 1.9.x release. Next release will make use of interfaces not available in 1.9.x.

For more details see commits.


Jaeyoung-Lim:
241 files changed!

PR #440, opened 2019-07-09
Add check for outsideYawFOV

nicovanduijn:
This is a quick fix that helps reduce pitch-oscillations produced by accelerating quickly such that the setpoint would then be outside the FOV.

Scaling according to pitch might be even better, but much more complicated and lead to overall slower flight. (And potentially very weird oscillations during acceleration)


nicovanduijn:
Honestly I think flight performance is unacceptable without #374 

Can someone else verify in SITL?

Issue #425, opened 2019-06-28
Global planner node-9 process died


JoeyLeeNPU:
When I use the global planner, I often meet a problem as following:

****
[global_planner_node-9] process has died [pid 5793, exit code -11, cmd /home/nb/avoidance_ws/devel/lib/global_planner/global_planner_node __name:=global_planner_node __log:=/home/nb/.ros/log/42cfa022-9983-11e9-ac6e-0242b966eb14/global_planner_node-9.log].
log file: /home/nb/.ros/log/42cfa022-9983-11e9-ac6e-0242b966eb14/global_planner_node-9*.log
****

The launch file is default, and the environment is Gazebo simulation.
Some parameters we set as "robot radius = 0.5", "resolution = 1.0", "cell scale = 1.0".
And another question is how should I set above parameters, and could tell me the meaning of "cell scale"? 

Waiting for response, thanks.

mrivi:
@Jaeyoung-Lim it's a know issue. We are trying to figure it out. If you have any insights, they're the most welcomed.

JoeyLeeNPU:
> @Jaeyoung-Lim it's a know issue. We are trying to figure it out. If you have any insights, they're the most welcomed.

Until now, we have tried to adjust these three parameters, we found that the frequency of "process died" is different(**when we set "robot radius = 0.5", "resolution = 1.0 or 0.5", "cell scale = 1.0", the number of died is smaller than other condition**), and this "died" frequency may also related positively to distance between current position("s" in code) and goal("t" in code).

We think that the problem may happen in **data format** because this global planner algorithm uses octomap and cell to represent discrete space and the drone.
Another is that during the path planning process, there are a mount of octomap grid information has been recorded, will it cause a problem such as memory crash?

Thanks so much for your response!

mrivi:
FYI @Jaeyoung-Lim

jkflying:
@JoeyLeeNPU Have you been able to catch the segfault in a debugger so we can get a line number? Maybe using GDB or Valgrind? I tried to catch the segfault but it wouldn't happen when I ran in the debugger.

JoeyLeeNPU:
I meet this problem when I run the global planner and this problem shows in the terminal. I copy the whole error message in my first description. And I don’t use other debugger. So how should I do next?


| |
李卓一 NPU
|
|
邮箱：15929972316@163.com
npulzy
|

签名由 网易邮箱大师 定制

On 07/01/2019 23:11, Julian Kent wrote:

[...]

PR #441, opened 2019-07-09
Fix for segfault with mutex

Jaeyoung-Lim:
This PR is an attempt to fix issue #425, where the global_planner has been crashing with a segfault.

The issue comes from the map update being done by deleting the `octree_` pointer while the replanning was still using the `octree_` information.


Jaeyoung-Lim:
Screenshot from 2019-07-10 10-49-51

PR #442, opened 2019-07-10
Small cleanups

nicovanduijn:
- Removes some line-wrapping in ROS prints that are no longer necessary
- Removes unused boolean in `getCostMatrix`


nicovanduijn:
rebased.
LGTM

PR #443, opened 2019-07-10
Simplify cost functions

nicovanduijn:
These are some pretty drastic (and for now, experimental) changes on the algorithmic side:

- We always plan to the sensor horizon (#374 )
- The tree generation as well as the "next node to add" decision are now based on the same cost function
- The "tree cost function" was removed completely
- The "histogramic cost function" was reduced in complexity by orders of magnitude. It now does basically only a decision based on "direction to goal" vs. "distance to obstacles". We should probably re-introduce one term that takes into account current velocity
- The tree heuristic was reduced in complexity as well, now just cartesian distance to goal (very common heuristic for A* path planning)

This brutally rips out tons of sophisticated logic, yet somehow it flies surprisingly well in SITL

### DIFF
--- a/avoidance/src/common.cpp
+++ b/avoidance/src/common.cpp
@@ -226,6 +226,11 @@ float wrapAngleToPlusMinusPI(float angle) { return angle - 2.0f * M_PI_F * std::
 
 float wrapAngleToPlusMinus180(float angle) { return angle - 360.f * std::floor(angle / 360.f + 0.5f); }
 
+float angleDifference(float a, float b) {
+  float angle = fmod(a - b, 360.f);
+  return angle >= 0.f ? (angle < 180.f) ? angle : angle - 360.f : (angle >= -180.f) ? angle : angle + 360.f;
+}
+
 double getAngularVelocity(float desired_yaw, float curr_yaw) {
   desired_yaw = wrapAngleToPlusMinusPI(desired_yaw);
   float yaw_vel1 = desired_yaw - curr_yaw;
--- a/local_planner/src/nodes/planner_functions.cpp
+++ b/local_planner/src/nodes/planner_functions.cpp
@@ -63,8 +63,7 @@ void processPointcloud(pcl::PointCloud<pcl::PointXYZI>& final_cloud,
 
         // only remember point if it's in a cell not previously populated by
         // complete_cloud, as well as outside FOV and 'young' enough
-        if (histogram_points_counter(p_ind.y(), p_ind.x()) < min_num_points_per_cell && xyzi.intensity < max_age &&
-            !pointInsideFOV(fov, p_pol_fcu)) {
+        if (xyzi.intensity < max_age && !pointInsideFOV(fov, p_pol_fcu)) {
           final_cloud.points.push_back(toXYZI(toEigen(xyzi), xyzi.intensity + elapsed_s));
 
           // to indicate that this cell now has a point
--- a/local_planner/src/nodes/planner_functions.cpp
+++ b/local_planner/src/nodes/planner_functions.cpp
@@ -15,16 +15,17 @@ void processPointcloud(pcl::PointCloud<pcl::PointXYZI>& final_cloud,
                        const std::vector<FOV>& fov, float yaw_fcu_frame_deg, float pitch_fcu_frame_deg,
                        const Eigen::Vector3f& position, float min_realsense_dist, float max_age, float elapsed_s,
                        int min_num_points_per_cell) {
+  const int SCALE_FACTOR = 4;
   pcl::PointCloud<pcl::PointXYZI> old_cloud;
   std::swap(final_cloud, old_cloud);
   final_cloud.points.clear();
   final_cloud.width = 0;
-  final_cloud.points.reserve((2 * GRID_LENGTH_Z) * (2 * GRID_LENGTH_E));
+  final_cloud.points.reserve((SCALE_FACTOR * GRID_LENGTH_Z) * (SCALE_FACTOR * GRID_LENGTH_E));
 
   float distance;
 
   // counter to keep track of how many points lie in a given cell
-  Eigen::MatrixXi histogram_points_counter(180 / (ALPHA_RES / 2), 360 / (ALPHA_RES / 2));
+  Eigen::MatrixXi histogram_points_counter(180 / (ALPHA_RES / SCALE_FACTOR), 360 / (ALPHA_RES / SCALE_FACTOR));
   histogram_points_counter.fill(0);
 
   for (const auto& cloud : complete_cloud) {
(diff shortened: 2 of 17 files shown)


local_planner/src/nodes/star_planner.cpp
```diff
@@ -144,7 +100,11 @@ void StarPlanner::buildLookAheadTree() {
...
           tree_.back().last_z_ = p_pol.z;
           float h = treeHeuristicFunction(tree_.size() - 1);
-          float c = treeCostFunction(tree_.size() - 1);
+          float distance_cost = 0.f;
+          float c = 0.f;
+          Eigen::Vector2i idx_ppol = polarToHistogramIndex(p_pol, ALPHA_RES);
+          float obstacle_distance = histogram.get_dist(idx_ppol.x(), idx_ppol.y());
+          costFunction(p_pol.e, p_pol.z, obstacle_distance, goal_, node_location, cost_params_, distance_cost, c);
           tree_.back().heuristic_ = h;
           tree_.back().total_cost_ = tree_[origin].total_cost_ - tree_[origin].heuristic_ + c + h;
```
jkflying: I'd prefer to see the `*0.1` scaling values over here, where we actually assemble the costs together. At some point we can make it something like a dot product between weights and a costs vectors.

local_planner/src/nodes/planner_functions.cpp
```diff
@@ -306,53 +304,20 @@ void padPolarMatrix(const Eigen::MatrixXf& matrix, unsigned int n_lines_padding,
...
-      cost_params.heading_cost_param * (projected_heading.topRows<2>() - projected_candidate.topRows<2>()).norm();
+  // goal cost
+  other_costs = (cost_params.heading_cost_param * (z_angle - facing_goal.z) * (z_angle - facing_goal.z) +
+                 cost_params.goal_cost_param * (e_angle - facing_goal.e) * (e_angle - facing_goal.e));
 
   // distance cost
-  distance_cost = 0.0f;
   if (obstacle_distance > 0.0f) {
-    distance_cost = 700.0f / obstacle_distance;
+    distance_cost = 100000.0f / std::exp(obstacle_distance);  // explodes around 5 meters
```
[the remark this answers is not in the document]
nicovanduijn: I like that idea! I'll look into it. Essentially we need to have a function that "explodes" around MPC_COL_PREV_D, that would be super nice! Only issue I see is that someone might fly with disabled collision prevention, but still expect the avoidance to work

local_planner/src/nodes/star_planner.cpp
```diff
@@ -144,7 +100,11 @@ void StarPlanner::buildLookAheadTree() {
           tree_.back().last_e_ = p_pol.e;
           tree_.back().last_z_ = p_pol.z;
           float h = treeHeuristicFunction(tree_.size() - 1);
-          float c = treeCostFunction(tree_.size() - 1);
+          float distance_cost = 0.f;
```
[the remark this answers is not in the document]
nicovanduijn: this has changed in my latest refactor

local_planner/src/nodes/star_planner.cpp
nicovanduijn: I took it out altogether

local_planner/src/nodes/planner_functions.cpp
```diff
@@ -305,54 +303,24 @@ void padPolarMatrix(const Eigen::MatrixXf& matrix, unsigned int n_lines_padding,
...
-  // combine costs
-  other_costs = 0.0f;
-  other_costs = yaw_cost + cost_params.height_change_cost_param_adapted * pitch_cost_up +
-                cost_params.height_change_cost_param * pitch_cost_down + yaw_cost_smooth + pitch_cost_smooth +
-                heading_cost;
+      cost_params.yaw_cost_param * (candidate_polar.z - facing_goal.z) * (candidate_polar.z - facing_goal.z);
+  float pitch_cost =
+      cost_params.pitch_cost_param * (candidate_polar.e - facing_goal.e) * (candidate_polar.e - facing_goal.e);
+  distance_cost =
+      obstacle_distance > 0 ? 2000.0f / (std::exp((obstacle_distance - cost_params.obstacle_cost_param))) : 0.0f;
```
jkflying: I still think a reversed sigmoid function would better represent the cost here.

Eg. 100/((1 + exp(obstacle_distance - cost_params.obstacle_cost_param))^2)

nicovanduijn: agreed, an upper bound is desirable

local_planner/src/nodes/planner_functions.cpp
```diff
@@ -305,54 +303,24 @@ void padPolarMatrix(const Eigen::MatrixXf& matrix, unsigned int n_lines_padding,
...
+  const float velocity_cost =
+      cost_params.velocity_cost_param * (velocity.norm() - candidate_velocity_cartesian.normalized().dot(velocity));
+  const float yaw_cost =
+      cost_params.yaw_cost_param * (candidate_polar.z - facing_goal.z) * (candidate_polar.z - facing_goal.z);
+  const float pitch_cost =
+      cost_params.pitch_cost_param * (candidate_polar.e - facing_goal.e) * (candidate_polar.e - facing_goal.e);
+  const float d = cost_params.obstacle_cost_param - obstacle_distance;
+  distance_cost = obstacle_distance > 0 ? 5000.0f * (1 + d / sqrt(1 + d * d)) : 0.0f;
+  other_costs = velocity_cost + yaw_cost + pitch_cost;
+  return velocity_cost + yaw_cost + pitch_cost + distance_cost;
```
jkflying: Just FYI, nobody is using the returned costs. maybe we should get rid of them?

nicovanduijn: They're used here: (and the following lines)
https://github.com/PX4/avoidance/blob/9782969475dd723c4d49532861e641fbd6ca47ee/local_planner/src/nodes/planner_functions.cpp#L152

However, I've been meaning to find a way to get rid of this. They're mostly used for visualization, which we don't need to compute for every node expansion of the A*. Would be a bigger refactor though

local_planner/src/nodes/planner_functions.cpp
```diff
@@ -63,8 +63,7 @@ void processPointcloud(pcl::PointCloud<pcl::PointXYZI>& final_cloud,
 
         // only remember point if it's in a cell not previously populated by
         // complete_cloud, as well as outside FOV and 'young' enough
-        if (histogram_points_counter(p_ind.y(), p_ind.x()) < min_num_points_per_cell && xyzi.intensity < max_age &&
```
nicovanduijn: We could try `histogram_points_counter(p_ind.y(), p_ind.x()) < 4 * min_num_points_per_cell`

jkflying: I think a better way would be to make the resolution of the downsampling grid 4x4 higher than the histogram grid, rather than 2x2 higher. This would allow points to get to 60 degrees off-center before any cells in the lower resolution grid would be empty upon return.

avoidance/src/common.cpp
```diff
@@ -226,6 +226,11 @@ float wrapAngleToPlusMinusPI(float angle) { return angle - 2.0f * M_PI_F * std::
 
 float wrapAngleToPlusMinus180(float angle) { return angle - 360.f * std::floor(angle / 360.f + 0.5f); }
 
+float angleDifference(float a, float b) {
+  float angle = a - b;
+  return std::abs(angle >= 0.f ? (angle < 180.f) ? angle : angle - 360.f : (angle >= -180.f) ? angle : angle + 360.f);
```
jkflying: Maybe we don't need to do an `abs` here, since you're squaring it down below?

(and then the doc string can change too)

avoidance/src/common.cpp
```diff
@@ -226,6 +226,11 @@ float wrapAngleToPlusMinusPI(float angle) { return angle - 2.0f * M_PI_F * std::
 
 float wrapAngleToPlusMinus180(float angle) { return angle - 360.f * std::floor(angle / 360.f + 0.5f); }
 
+float angleDifference(float a, float b) {
+  float angle = a - b;
```
jkflying: To be safe, rather do 
```suggestion
  float angle = fmod(a - b, 360.f);
```

nicovanduijn: done

avoidance/src/common.cpp
nicovanduijn: done

PR #448, opened 2019-07-22
Refactor getDirectionFromTree

nicovanduijn:
This PR refactors the planner function `getDirectionFromTree`.

The old version used a complicated method to determine which tree segment we are currently on and then interpolated between the nodes to find a setpoint. This logic is not adequate for some corner cases, when the planned tree yielded an endpoint that is actually closer to the drone than the start node, it would switch to `goStraight` mode and fly directly into an obstacle.
The interpolation also doesn't make much sense, since the output is anyway taken as a unit vector which is later scaled depending on FOV and desired velocity.

The new logic simply takes the current velocity estimate and the tree generation time to determine which tree segment we are currently on, and then returns the next node in the tree as the setpoint.
The waypoint generator creates a unit vector toward this location to put into the `goto_position` to comply with the rest of the logic in the `waypoint_generator`.

### DIFF
--- a/local_planner/include/local_planner/planner_functions.h
+++ b/local_planner/include/local_planner/planner_functions.h
@@ -148,7 +148,7 @@ void printHistogram(const Histogram& histogram);
 
 /**
 * @brief      Returns a setpoint that lies on the given path
-* @param[in]  vector of nodes defining the path
+* @param[in]  vector of nodes defining the path, with the last node of the path at index 0
 * @param[in]  ros time of path generation
 * @param[in]  velocity, scalar value for the norm of the current vehicle velocity
 * @param[out] setpoint on the tree toward which the drone should fly
--- a/local_planner/include/local_planner/planner_functions.h
+++ b/local_planner/include/local_planner/planner_functions.h
@@ -147,14 +147,14 @@ Eigen::ArrayXf getConicKernel(int radius);
 void printHistogram(const Histogram& histogram);
 
 /**
-* @brief      finds the minimum cost direction in the tree
-* @param[in]  vector of nodes defining the tree
-* @param[in]  ros time of tree generation
-* @param[in]  velocity, scalar value for the current vehicle velocitz
+* @brief      Returns a setpoint that lies on the given path
+* @param[in]  vector of nodes defining the path
+* @param[in]  ros time of path generation
+* @param[in]  velocity, scalar value for the norm of the current vehicle velocity
 * @param[out] setpoint on the tree toward which the drone should fly
 * @returns    boolean indicating whether the tree was valid
 **/
-bool getSetpointFromTree(const std::vector<Eigen::Vector3f>& tree, const ros::Time& tree_generation_time,
+bool getSetpointFromPath(const std::vector<Eigen::Vector3f>& path, const ros::Time& path_generation_time,
                          float velocity, Eigen::Vector3f& setpoint);
 }
 #endif  // LOCAL_PLANNER_FUNCTIONS_H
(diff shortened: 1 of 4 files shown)


local_planner/src/nodes/planner_functions.cpp
```diff
@@ -355,74 +355,25 @@ void costFunction(float e_angle, float z_angle, float obstacle_distance, const E
...
-    tree_available = false;
+  // step through the tree until the point where we should be if we had traveled perfectly with velocity along it
+  float distance_left = std::max(0.1, (ros::Time::now() - tree_generation_time).toSec() * velocity);
+  setpoint = tree[i - 2];
+  Eigen::Vector3f tree_segment = tree[i - 2] - tree[i - 1];
+
+  for (i = tree.size() - 2; i > 0 && distance_left > tree_segment.norm(); --i) {
+    distance_left -= tree_segment.norm();
+    tree_segment = tree[i - 1] - tree[i];
+    setpoint = tree[i - 1];
```
jkflying: I think we should do a linear interpolation for the final points, otherwise we get sharp jumps in the setpoint

baumanta:
Does this PR include planning all the time or does the go straight mode still exist?  If go straight still exists how do you decide whether the current tree is valid or not? 

E.g. If we have followed a nice path to its end, we do not want to stop at the last node but continue to go straight. Whereas if we get such a zigg zagg path and reach the last node (because it is basically equal to the first node) we do not want to go straight but stay at the last node.

local_planner/src/nodes/planner_functions.cpp
nicovanduijn: now we're sending a setpoint one segment ahead, which should remove those jumps

nicovanduijn:
You're right, that logic is almost dead now. The only way it will still switch to 'goStraight' is if the tree is invalid. (Too short). This is implicitly done in the local_planner by not computing a tree at all if the histogram is empty.

I've been trying to remove the 'goStraight' altogether, but my problem right now is that this removes the option to return to the mission line as well. I'd have to move that part to the planner

nicovanduijn:
I've found a clever way around the issue @baumanta pointed out:
By returning `i > 0` in `getSetpointFromTree` we achieve the desired result: a tree will be invalid if we've flown past it, and it will switch to `goStraight` nicely

local_planner/src/nodes/planner_functions.cpp
```diff
@@ -355,74 +355,32 @@ void costFunction(float e_angle, float z_angle, float obstacle_distance, const E
...
-        second_min_dist_idx = min_dist_idx;
-        second_min_dist = min_dist;
-        min_dist = distances[i];
-        min_dist_idx = i;
-      } else if (distances[i] < second_min_dist) {
-        second_min_dist = distances[i];
-        second_min_dist_idx = i;
-      }
-    }
+bool getSetpointFromTree(const std::vector<Eigen::Vector3f>& tree, const ros::Time& tree_generation_time,
```
baumanta: I think naming the argument `tree` is rather confusing. The tree is the container of all expanded nodes, whereas what you actually put in is the precomputed path (list of nodes) through the tree. Would opt for `path` or similar

nicovanduijn: I'll rename to `getSetpointFromPath` and the variable to `path`

local_planner/include/local_planner/planner_functions.h
```diff
@@ -147,12 +147,14 @@ Eigen::ArrayXf getConicKernel(int radius);
 void printHistogram(const Histogram& histogram);
 
 /**
-* @brief      finds the minimum cost direction in the tree
-* @param[out] p_pol, polar coordinates of the cheapest direction
-* @param[in]  path_node_positions, array of expanded tree nodes
-* @param[in]  position, current vehicle position
+* @brief      Returns a setpoint that lies on the given path
+* @param[in]  vector of nodes defining the path
```
jkflying: Probably mention that the last node should have index 0

nicovanduijn: done

PR #450, opened 2019-07-23
Remove ground cropping

nicovanduijn:
This removes the ground cropping


local_planner/include/local_planner/local_planner.h
```diff
@@ -120,7 +120,7 @@ class LocalPlanner {
   void generateHistogramImage(Histogram& histogram);
 
  public:
-  Box histogram_box_;
+  //Box histogram_box_;
```
jkflying: Delete, don't comment (and ground_distance below)

local_planner/src/nodes/local_planner.cpp
```diff
@@ -167,7 +166,7 @@ void LocalPlanner::updateObstacleDistanceMsg(Histogram hist) {
   msg.header.frame_id = "local_origin";
   msg.angle_increment = static_cast<double>(ALPHA_RES) * M_PI / 180.0;
   msg.range_min = min_realsense_dist_;
-  msg.range_max = histogram_box_.radius_;
+  msg.range_max = sensor_range_;//histogram_box_.radius_;
```
jkflying: delete comment here too

local_planner/src/nodes/planner_functions.cpp
```diff
@@ -32,9 +31,8 @@ void processPointcloud(pcl::PointCloud<pcl::PointXYZI>& final_cloud,
     for (const pcl::PointXYZ& xyz : cloud) {
       // Check if the point is invalid
       if (!std::isnan(xyz.x) && !std::isnan(xyz.y) && !std::isnan(xyz.z)) {
-        if (histogram_box.isPointWithinBox(xyz.x, xyz.y, xyz.z)) {
           distance = (position - toEigen(xyz)).norm();
-          if (distance > min_realsense_dist && distance < histogram_box.radius_) {
+          if (distance > min_realsense_dist ) {
```
jkflying:
```suggestion
          if (distance > min_realsense_dist && distance < sensor_range ) {
```

And you need to bring in the sensor_range from the planner

local_planner/src/nodes/planner_functions.cpp
```diff
@@ -52,9 +49,8 @@ void processPointcloud(pcl::PointCloud<pcl::PointXYZI>& final_cloud,
   for (const pcl::PointXYZI& xyzi : old_cloud) {
     // adding older points if not expired and space is free according to new
     // cloud
-    if (histogram_box.isPointWithinBox(xyzi.x, xyzi.y, xyzi.z)) {
       distance = (position - toEigen(xyzi)).norm();
-      if (distance < histogram_box.radius_) {
+
```
jkflying:
```suggestion
if (distance < sensor_range)
```

local_planner/src/nodes/planner_functions.cpp
nicovanduijn: For the memory points - sure we need to make sure it doesn't blow up. But I don't see why we would arbitrarily limit the range of our sensor. This way, as far as you see, you have data. Why the hell not. Makes the algorithm sensor-independent. Ignoring points that are far away is done by the obstacle cost term in the cost function.

local_planner/src/nodes/planner_functions.cpp
```diff
@@ -50,28 +46,23 @@ void processPointcloud(pcl::PointCloud<pcl::PointXYZI>& final_cloud,
...
-        if (histogram_points_counter(p_ind.y(), p_ind.x()) < min_num_points_per_cell && xyzi.intensity < max_age &&
-            !pointInsideFOV(fov, p_pol_fcu)) {
-          final_cloud.points.push_back(toXYZI(toEigen(xyzi), xyzi.intensity + elapsed_s));
-
-          // to indicate that this cell now has a point
-          histogram_points_counter(p_ind.y(), p_ind.x()) = min_num_points_per_cell;
-        }
-      }
+    // adding older points if not expired and space is free according to new cloud
+    distance = (position - toEigen(xyzi)).norm();
```
nicovanduijn: unused

baumanta:
did not look at it in detail, but I'm super glad if we don't need this ugly piece of code anymore! :)

nicovanduijn:
We should probably flight-test this a few times, since it will significantly increase the size of the local point cloud.
I doubt it will have an effect on overall performance, since we're still talking about small point clouds (downsampled!) and far from the memory constraints of our system, but it's still worth a check.

local_planner/src/nodes/local_planner.cpp
```diff
@@ -28,7 +28,8 @@ void LocalPlanner::setState(const Eigen::Vector3f& pos, const Eigen::Vector3f& v
 
 // set parameters changed by dynamic rconfigure
 void LocalPlanner::dynamicReconfigureSetParams(avoidance::LocalPlannerNodeConfig& config, uint32_t level) {
-  histogram_box_.radius_ = static_cast<float>(config.box_radius_);
+  //  histogram_box_.radius_ = static_cast<float>(config.box_radius_);
```
baumanta: I think this comment was missed when deleting old stuff

jkflying:
This flew well, but seemed to keep very high altitude in certain cases, being unable to hit setpoints because it couldn't descend enough.

This will also need some reworking of the collision prevention laser scan generation to ignore points below a certain distance below, since there is no distance cropping anymore and points that are very far away but within the 30 degree VFOV trigger collision warnings.

jkflying:
Flight tests are showing a major performance bottleneck with this PR, planner is dropping to ~3hz which also affects the laser scan used for collision prevention.

I'll back out the sensor range changes, and keep this just at ground cropping removal.

baumanta:
Sounds good to me, maybe we should do one more test flight to see whether the collision prevention still works nicely at low altitude and the frequency is fine again.

jkflying:
Did another flight test and it seemed really poor performance in collision prevention. Often it didn't sense objects until very late.
https://logs.px4.io/plot_app?log=7e524c35-4fa2-42bf-9b73-0cd7555f8407

Will investigate further.

PR #451, opened 2019-07-26
Fix tf exception

Jaeyoung-Lim:
After #430, `global_planner` fails with the following error.
```
terminate called after throwing an instance of 'tf2::LookupException'
  what():  "map" passed to lookupTransform argument source_frame does not exist.
```

This PR gets rid of unneeded tf listeners to prevent this happening again. The error was introduced as mavros config changed its `frame_id`s for some frames.

### DIFF
--- a/global_planner/include/global_planner/global_planner_node.h
+++ b/global_planner/include/global_planner/global_planner_node.h
@@ -116,6 +116,7 @@ class GlobalPlannerNode {
   double speed_ = min_speed_;
   double start_yaw_;
   bool position_received_;
+  std::string frame_id_;
 
   // Dynamic Reconfiguration
   double clicked_goal_alt_;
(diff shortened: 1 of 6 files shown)


nicovanduijn:
looks like a no-brainer.
LGTM

Jaeyoung-Lim:
@nicovanduijn Can we merge this since `global_planner` is broken in master

mrivi:
@Jaeyoung-Lim is the `world` frame used anywhere in the global planner? Otherwise we can also remove https://github.com/PX4/avoidance/blob/master/global_planner/launch/global_planner_depth-camera.launch#L12-L13

Jaeyoung-Lim:
@mrivi Maybe no longer after this PR. I will do a check

Jaeyoung-Lim:
@mrivi Removed the static tf publishers!

Issue #132, opened 2018-11-02
[Feature] Plug OMPL to Avoidance to open the possibility of using other motion planning algorithms


TSC21:
[OMPL](http://ompl.kavrakilab.org/) (Open Motion Planning Library) is a library which contains most of the state-of-the-art motion and path planning algorithms. It's already used in various robotics frameworks, including MoveIt!. Would be interesting to explore the possibility of integrate it with the Avoidance API in order to add other alternatives the global planner.
Note also that the [license](http://ompl.kavrakilab.org/license.html) is permissive and should set us good to go on the integration.
I am open to discuss this on the next avoidance call.
@mrivi, @baumanta, @nicovanduijn what do you think?

TSC21:
Note that it can also be installed directly from the ROS deb packages using `sudo apt-get install ros-`rosversion -d`-ompl`.

Jaeyoung-Lim:
I would be excited to see this moving forward

TSC21:
@Jaeyoung-Lim is this something you have the bandwidth to explore?

ytaskiran:
What happened to this project? It seemed some progress had taken place at the time but I don't understand where it went.

Jaeyoung-Lim:
@ytaskiran This is how far it went: https://github.com/PX4/PX4-Avoidance/pull/453

LorenzMeier:
@ytaskiran Would you want to take it on from there?

ytaskiran:
@LorenzMeier Actually I'd love to.  By the way, I think moving on with ROS2 instead of mavros would be more convenient from now on. What do you think?

PR #453, opened 2019-07-29
Use OMPL for global path planning for global_planner

Jaeyoung-Lim:
This PR integrates [Open Motion Planning Library](https://ompl.kavrakilab.org/) into global_planner. 

Most of the previous `global_planner` has been written from the ground up, making it quite hard to modify the planning algorithm itself.

Having OMPL as a planner backend will give more versatility to add / modify the path planning algorithm and give access to the various planners that are available in OMPL. Moreover, given that ompl do not depend on map representations, this could potentially enable the ability to use different map representations for different use cases. 

A lot have the structure has been used following https://github.com/ethz-asl/mav_voxblox_planning.

Currently depends on #452 and will rebase once merged. Was first proposed in #132

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ before_install:
   # Install yaml-cpp ROS package
   - sudo apt-get install -y libyaml-cpp-dev
   # Install MavLink and Mavros
-  - sudo apt-get install -y ros-$ROS_DISTRO-mavlink ros-$ROS_DISTRO-mavros-msgs ros-$ROS_DISTRO-mavros ros-$ROS_DISTRO-mavros-extras
+  - sudo apt-get install -y ros-$ROS_DISTRO-mavlink ros-$ROS_DISTRO-mavros-msgs ros-$ROS_DISTRO-mavros ros-$ROS_DISTRO-mavros-extras ros-$ROS_DISTRO-ompl
   # Source environment
   - source /opt/ros/$ROS_DISTRO/setup.bash
   # Prepare rosdep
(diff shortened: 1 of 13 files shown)


mamoll:
Nice work! I am one of the OMPL developers. I see that RRT* is made the default planner. Is this the fastest in your testing? Note that we have added a [CostConvergenceTerminationCondition class](http://ompl-beta.kavrakilab.org/classompl_1_1base_1_1CostConvergenceTerminationCondition.html) that you can use to have RRT* terminate upon convergence rather than wait for a timeout. Running a feasible planner (e.g., RRTConnect) followed by path simplification often works "well enough" and can be much faster, so make sure to give that a try, too. 

Finally, you may also want to experiment with different state spaces. I can imagine that high-speed drone flight kinodynamic constraints become very important. OMPL has several kinodynamic planners that, once you implement a steering function, could also work well. A reasonable approximation of the kinodynamic constraints can be obtained by modeling the drone as a Dubins vehicle or Reeds-Sheep vehicle with translation in the Z direction. We have successfully used this in a collaboration with the University of Girona for AUVs. See [[1]](https://moll.ai/publications/vidal2019online-multilayered-motion-planning.pdf) and [[2]](https://moll.ai/publications/hernandez2019online-motion-planning-for-unexplored.pdf). This can be done by creating your own custom StateSpace (e.g., a CompoundStateSpace comprised of DubinsStateSpace and a RealVectorStateSpace).

Jaeyoung-Lim:
@mamoll Thanks for the insights and pointers! I am using RRT* simply as a example. There is a lot to improve, but the current state I would be satisfied would be to have a path searching global planner(Such as RRT*) with polynomial path smoothing.

I would definitely want to move towards the planning with dynamic constraints. This is becoming more and more of an issue not just for the dynamic constraints but also for the limited sensor range vehicles normally have.

However, most of the leg work that needs to be done is integrating ompl with the PX4 flight stack.

PR #454, opened 2019-07-29
Deprecate launchfile generation

Jaeyoung-Lim:
Launchfile generation will not be maintained as it will be usecase specific, and will be hard to have a general launchfile generation script.

A better way would be to have example launchfiles that can be used for real vehicles. 

This PR deprecates the launchfile generation scripts


SalimTerryLi:
Hi
It seems old readme still refers to this script. Is there any documents about how to create one from the stretch?

PR #457, opened 2019-08-05
SLP: grid exploration

mrivi:
if the area below the vehicle isn't good enough to land, we examine the entire grid with starting from the center with outward moving spiral sliding window.
if the Grid evaluation finds a suitable spot to land the splanner switches into goto state to go to that point and then directly lands there without re-evaluationg the area.
if the grid evaluation doesn't find any landing area the planner switches into GoTo to perform the spiral patter to find a landing location outside the current grid


safe_landing_planner/src/nodes/waypoint_generator.cpp
```diff
@@ -66,13 +88,16 @@ void WaypointGenerator::updateSLPState() {
...
 SLPState WaypointGenerator::chooseNextState(SLPState currentState, usm::Transition transition) {
   prev_slp_state_ = currentState;
-  USM_TABLE(currentState, SLPState::GOTO,
-            USM_STATE(transition, SLPState::GOTO, USM_MAP(usm::Transition::NEXT1, SLPState::ALTITUDE_CHANGE);
-                      USM_MAP(usm::Transition::NEXT2, SLPState::LOITER));
-            USM_STATE(transition, SLPState::ALTITUDE_CHANGE, USM_MAP(usm::Transition::NEXT1, SLPState::LOITER));
-            USM_STATE(transition, SLPState::LOITER, USM_MAP(usm::Transition::NEXT1, SLPState::LAND);
-                      USM_MAP(usm::Transition::NEXT2, SLPState::GOTO));
-            USM_STATE(transition, SLPState::LAND, ));
+  USM_TABLE(
```
jkflying: clang-format is really ugly here. I think turn it off with `// clang-format off` (and on again below with `// clang-format on`) and then make this a bit nicer, so that it is easier to read. One USM_MAP on each line, etc

safe_landing_planner/src/nodes/waypoint_generator.cpp
```diff
@@ -102,15 +130,25 @@ usm::Transition WaypointGenerator::runGoTo() {
...
            velocity_setpoint_.x(), velocity_setpoint_.y(), velocity_setpoint_.z());
   altitude_landing_area_percentile_ = landingAreaHeightPercentile(80.f);
-  std::fill(can_land_hysteresis_.begin(), can_land_hysteresis_.end(), 0.f);
+  can_land_hysteresis_matrix_.fill(0.0f);
 
   ROS_INFO("[WGN] Landing Radius: xy  %f, z %f ", (goal_.topRows<2>() - position_.topRows<2>()).norm(),
            fabsf(position_.z() - altitude_landing_area_percentile_));
 
+  if (found_land_area_in_grid_) {
+    landing_radius_ = 0.5f;
```
jkflying: Is this correct? Do we always want to use 0.5?

safe_landing_planner/src/nodes/waypoint_generator.cpp
```diff
@@ -221,6 +262,46 @@ usm::Transition WaypointGenerator::runLand() {
...
+    }
+  }
+  start_grid_exploration_ = false;
+  return usm::Transition::NEXT2;
+}
+
+bool WaypointGenerator::evaluatePatch(Eigen::Vector2i &left_upper_corner) {
+  Eigen::MatrixXi kernel(can_land_hysteresis_matrix_.rows(), can_land_hysteresis_matrix_.cols());
+  kernel.fill(0);
+  kernel.block(left_upper_corner.x(), left_upper_corner.y(), mask_.rows(), mask_.cols()) = mask_;
```
jkflying: Careful with this. Block indexing is row, column which corresponds to y, x

safe_landing_planner/src/nodes/waypoint_generator.cpp
mrivi: When doing the grid evaluation and spiral pattern we need to be much more strict than with normal waypoints. The default acceptance radius is 2m. How would you feel about having it on a different variable?

mrivi:
new_state_machine

Issue #458, opened 2019-08-05
Real flight not possible due to MAVROS configs


nicovanduijn:
**Describe the bug**
Since moving the MAVROS configs to be maintained upstream (https://github.com/PX4/avoidance/pull/430), mavros does no longer publish the required transforms.

This was done previously in the (now unused) [config file](https://github.com/PX4/avoidance/blob/master/local_planner/resource/px4_config.yaml)  in the `local_position` section.

In SITL it still works because the odom plugin seems to be running, which publishes a transform 
from `local_origin_ned` to `fcu_frd` [here]( https://github.com/mavlink/mavros/blob/8aabb0bf1b7ba05ddd5da487bbb729655a37a2e9/mavros_extras/src/plugins/odom.cpp#L215)

**Workaround**
Since the [mavros config file]((https://github.com/PX4/avoidance/blob/master/local_planner/resource/px4_config.yaml) ) was never actually removed, we can just use it again.

**Solution** 
Longer term, all `local_origin`  frames should be changed to `map` (convention upstream mavros) and the default for the `local_position` plugin should be to publish [here](https://github.com/mavlink/mavros/blob/8aabb0bf1b7ba05ddd5da487bbb729655a37a2e9/mavros/launch/px4_config.yaml#L71) and the SITL setup should make sure the ODOM plugin doesn't run, such that SITL breaks as well. Right now, the source for pose estimate is different in SITL (odometry plugin) than on the real vehicle (local_position plugin).

PR #459, opened 2019-08-05
Revert "Create common px4 config for avoidance (#430)"

jkflying:
This reverts commit ff9df729d59afdb8fc6620ac17be9330e5c7c989.

Quick fix for https://github.com/PX4/avoidance/issues/458

@Jaeyoung-Lim


Jaeyoung-Lim:
@jkflying In that case, could we at least make it use the common config in the avoidance package rather than `local_planner`? The main intention of #430 was not to use the configs from `mavros`, but to have it under the `avoidance` package.

jkflying:
Let's merge this first, then we can do that.

PR #467, opened 2019-08-14
Add window model

mrivi:
- Adds window model
- Fix frame for rviz visualization in boxes* worlds


jkflying:
It would be nice to have wall around the window instead of just the frame

mrivi:
wall

PR #468, opened 2019-08-16
Decelerate before altitude change

mrivi:
This PR makes enforces a low horizontal velocity before changing altitude. Fixes overshooting problem when landing


LorenzMeier:
I am starting to see more and more logic duplication with the core flight control software. Could you please discuss with @bresch a roadmap how you can get motion constraints from the flight controller?

We have been discussing this for a while as I saw the need early, but I think it is time now to have a very clear roadmap on this.

mrivi:
@LorenzMeier we're actually trying to solve the opposite problem with this PR. When a waypoint is blocked we cannot use the motion constraints of the flight controller because it isn't aware of waypoints being unreachable.
However I do generally agree that we have to much logic duplication with the Flight Controller and we should minimize that as much as possible.

PR #469, opened 2019-08-16
add fancy marker for range scan

baumanta:


mhkabir:
@baumanta would you be able to post a screenshot of what this looks like? :)

baumanta:
range_scan

baumanta:
@mhkabir  here you go :) It is just a visualization of the range scan message, which basically is a 1D histogram compression. This goes to the FCU as OBSTACLE_DISTANCE message and is used for collision prevention.

mhkabir:
Thanks, this looks great!

PR #472, opened 2019-08-26
Port FSM to Local Planner

mrivi:
The PR removes the offtrack logic. This logic needs to be baked into the cost function
local_planner_fsm


mrivi:
The last commit adds a new cost that pulls the vehicle towards the line between the previous and current waypoint. Depending on the weight of this yaw cost versus the goal yaw cost the vehicle is more or less agressive in re-entering on the line.

yaw_to_line_cost and yaw_cost weighted 50, 50
offtrack2

yaw_to_line_cost and yaw_cost weighted 25, 75
offtrack


I went for the 25, 75 weighting

mrivi:
Flight log https://review.px4.io/plot_app?log=62bda922-6280-4f37-b8ba-f5f136131f55

PR #474, opened 2019-08-27
Port LocalPlannerNode to nodelets

Jaeyoung-Lim:
This PR ports local planner to be able to run as [nodelets](http://wiki.ros.org/nodelet)

Nodelets enables us to run local_planner in platforms which has limited computational resources. The computational benefits come from Nodelets not serializing/deserializing pointclouds to a TCP protocol but sharing the messages via a shared_ptr.

An executable is also added as a node so that the planner can still run as a standalone node. @mrivi


Jaeyoung-Lim:
@mrivi Could you make a PR on this branch so I can start fixing the tests?

jkflying:
Deprecated in favor of #478

PR #483, opened 2019-09-13
Move config to avoidance package

Jaeyoung-Lim:
This PR moves the px4 configuration file to the avoidance package.

This fixes the problem where local_planner is required to be built to be able to run either the `safe_landing_planner` or the `local_planner`

This was brought up at #430 but introduced a lot of problems when using the configs from mavros. This is a follow up of #459 

Tested in SITL


bys1123:
image
Is there three files forgot to change?

Jaeyoung-Lim:
@bys1123 That looks like it. Could you make a PR for this?

Issue #486, opened 2019-09-18
issue when run the global_planner


onionysy:
Thank for the excellent 'avoidance' project.
i have run the local_planner successfully, but when i run `rosrun mavros mavsafety arm`, the cammand shows:
`coamc@coamc-HP-Z840-Workstation:~$ rosrun mavros mavsafety arm
Request failed. Check mavros logs`
and
`WARN  [commander] Arming denied! Avoidance system not ready
[ERROR] [1568781066.666805554, 39.476000000]: FCU: Arming denied! Avoidance system not ready`

can you give me some advices?
thank you!

mrivi:
@onionysy I can reproduce the problem, thanks for reporting. We'll look into it

mrivi:
fixed in #488

PR #434, opened 2019-07-05
Firmware Parameters in AvoidanceNode

mrivi:
- poll PX4 parameters in Avoidance Node
- use parameters in SafeLandingPlanner

TODO: - safe landing planner companion process status on avoidanceNode

PR #488, opened 2019-09-19
Fix global planner not arming

Jaeyoung-Lim:
This PR fixes the problem raised by #486 where the `global_planner` is not able to arm.

The regression was caused by #434 as the implementation requires the `avoidance_node` to be explicitly initialized.

Tested in SITL


kimzheng1204:
Does it have any differences with the original code? I find that the code you changed is the same as what I downloaded from the website.

PR #489, opened 2019-09-24
Use Eigen instead of ros_msgs for internal data passing

Jaeyoung-Lim:
A lot of the data that were using `geometry_msgs::PoseStamped` or `geometry_msgs::TwistStamped` were using only the position or linear velocity field in the messages. 

This PR refactors local_planner_node, where ros `msg` formats were used for internal data processing. the ros `msg` format is used only during the callback of the message and when publishing the message itself. 

- Improves the readability of the code as we don't need to look for data inside the messages
- Removes unnecessary calls to `toEigen` or `toPoint` frequently for internal data passing

Tested in SITL

### DIFF
--- a/local_planner/src/nodes/local_planner_node.cpp
+++ b/local_planner/src/nodes/local_planner_node.cpp
@@ -277,10 +277,9 @@ MAV_STATE LocalPlannerNode::getSystemStatus() { return avoidance_node_->getSyste
 void LocalPlannerNode::calculateWaypoints(bool hover) {
   bool is_airborne = armed_ && (nav_state_ != NavigationState::none);
 
-  wp_generator_->updateState(newest_position_, newest_orientation_,
-                             goal_position_, prev_goal_position_,
-                             velocity_, hover, is_airborne, nav_state_, is_land_waypoint_,
-                             is_takeoff_waypoint_, desired_velocity_);
+  wp_generator_->updateState(newest_position_, newest_orientation_, goal_position_, prev_goal_position_, velocity_,
+                             hover, is_airborne, nav_state_, is_land_waypoint_, is_takeoff_waypoint_,
+                             desired_velocity_);
   waypointResult result = wp_generator_->getWaypoints();
 
   Eigen::Vector3f closest_pt = Eigen::Vector3f(NAN, NAN, NAN);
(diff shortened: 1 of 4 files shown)


local_planner/src/nodes/local_planner_node.cpp
```diff
@@ -87,17 +87,18 @@ void LocalPlannerNode::startNode() {
 
 void LocalPlannerNode::readParams() {
   // Parameter from launch file
-  auto goal = toPoint(local_planner_->getGoal());
-  nh_.param<double>("goal_x_param", goal.x, 9.0);
-  nh_.param<double>("goal_y_param", goal.y, 13.0);
-  nh_.param<double>("goal_z_param", goal.z, 3.5);
+  float init_goal_position_x, init_goal_position_y, init_goal_position_z;
+
+  nh_.param<float>("goal_x_param", init_goal_position_x, 9.0);
```
jkflying: This would be neater as just:
```suggestion
  nh_.param<float>("goal_x_param", goal_position_.x(), 9.0);
```
For x,y,z
Then you don't need the temporary variables.

local_planner/src/nodes/local_planner_node.cpp
```diff
@@ -273,28 +277,26 @@ MAV_STATE LocalPlannerNode::getSystemStatus() { return avoidance_node_->getSyste
 void LocalPlannerNode::calculateWaypoints(bool hover) {
   bool is_airborne = armed_ && (nav_state_ != NavigationState::none);
 
-  wp_generator_->updateState(toEigen(newest_pose_.pose.position), toEigen(newest_pose_.pose.orientation),
-                             toEigen(goal_msg_.pose.position), toEigen(prev_goal_.pose.position),
-                             toEigen(vel_msg_.twist.linear), hover, is_airborne, nav_state_, is_land_waypoint_,
-                             is_takeoff_waypoint_, toEigen(desired_vel_msg_.twist.linear));
+  wp_generator_->updateState(newest_position_, newest_orientation_, goal_position_, prev_goal_position_, velocity_,
```
jkflying: :heart:

local_planner/src/nodes/local_planner_node.cpp
```diff
@@ -313,17 +315,17 @@ void LocalPlannerNode::clickedPointCallback(const geometry_msgs::PointStamped& m
...
 void LocalPlannerNode::clickedGoalCallback(const geometry_msgs::PoseStamped& msg) {
   new_goal_ = true;
-  prev_goal_ = goal_msg_;
-  goal_msg_ = msg;
+  prev_goal_position_ = goal_position_;
+  goal_position_ = toEigen(msg.pose.position);
   /* Selecting the goal from Rviz sets x and y. Get the z coordinate set in
    * the launch file */
-  goal_msg_.pose.position.z = local_planner_->getGoal().z();
+  goal_position_(2) = local_planner_->getGoal().z();
```
jkflying:
```suggestion
  goal_position_.z() = local_planner_->getGoal().z();
```

Issue #507, opened 2019-10-15
TF Buffer warnings during regular operation in SITL


jkflying:
**Describe the bug**
During SITL we see:
```
[ WARN] [1571146245.985945095, 162.456000000]: TF Buffer: could not retrieve requested transform from buffer, tf has not yet arrived
[ WARN] [1571146245.989237024, 162.460000000]: TF Buffer: could not retrieve requested transform from buffer, tf has not yet arrived
[ WARN] [1571146245.989300188, 162.460000000]: TF Buffer: could not retrieve requested transform from buffer, tf has not yet arrived
[ WARN] [1571146245.990096110, 162.460000000]: TF Buffer: could not retrieve requested transform from buffer, tf has not yet arrived
[ WARN] [1571146245.993266281, 162.464000000]: TF Buffer: could not retrieve requested transform from buffer, tf has not yet arrived
[ WARN] [1571146245.993519107, 162.464000000]: TF Buffer: could not retrieve requested transform from buffer, tf has not yet arrived
[ WARN] [1571146245.993681721, 162.464000000]: TF Buffer: could not retrieve requested transform from buffer, tf has not yet arrived
[ WARN] [1571146246.070028581, 162.540000000]: TF Buffer: could not retrieve requested transform from buffer, tf has not yet arrived
[ WARN] [1571146246.070572289, 162.540000000]: TF Buffer: could not retrieve requested transform from buffer, tf has not yet arrived
[ WARN] [1571146246.071190241, 162.540000000]: TF Buffer: could not retrieve requested transform from buffer, tf has not yet arrived
[ WARN] [1571146246.073358473, 162.544000000]: TF Buffer: could not retrieve requested transform from buffer, tf has not yet arrived
[ WARN] [1571146246.073424075, 162.544000000]: TF Buffer: could not retrieve requested transform from buffer, tf has not yet arrived
[ WARN] [1571146246.073815361, 162.544000000]: TF Buffer: could not retrieve requested transform from buffer, tf has not yet arrived
[ WARN] [1571146246.077698009, 162.548000000]: TF Buffer: could not retrieve requested transform from buffer, tf has not yet arrived
[ WARN] [1571146246.077774700, 162.548000000]: TF Buffer: could not retrieve requested transform from buffer, tf has not yet arrived
[ WARN] [1571146246.078085380, 162.548000000]: TF Buffer: could not retrieve requested transform from buffer, tf has not yet arrived
[...]

PR #511, opened 2019-10-17
Get Rid of TF Warnings

mrivi:
Fix for  #507

This PR makes sure that we wait for a pointcloud to be transformed before overwriting it with a new one. However if the pointcloud cannot be transformed for 1 second we drop it and use the newly arrived pointcloud.

### DIFF
--- a/local_planner/src/nodes/local_planner_nodelet.cpp
+++ b/local_planner/src/nodes/local_planner_nodelet.cpp
@@ -184,7 +184,6 @@ void LocalPlannerNodelet::updatePlannerInfo() {
     std::lock_guard<std::mutex> transformed_cloud_guard(*(cameras_[i].transformed_cloud_mutex_));
     try {
       local_planner_->original_cloud_vector_.push_back(std::move(cameras_[i].pcl_cloud));
-      cameras_[i].transformed_ = false;
       local_planner_->setFOV(i, cameras_[i].fov_fcu_frame_);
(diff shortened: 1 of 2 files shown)


local_planner/src/nodes/local_planner_nodelet.cpp
```diff
@@ -427,17 +426,24 @@ void LocalPlannerNodelet::printPointInfo(double x, double y, double z) {
...
-  if (!cameras_[index].transform_registered_) {
-    std::pair<std::string, std::string> transform_frames;
-    transform_frames.first = msg->header.frame_id;
-    transform_frames.second = "/local_origin";
-    buffered_transforms_.push_back(transform_frames);
-    cameras_[index].transform_registered_ = true;
+  float time_since_last_cloud_update =
+      fabsf(cameras_[index].newest_cloud_msg_.header.stamp.toSec() - msg->header.stamp.toSec());
+  if (cameras_[index].transformed_ || time_since_last_cloud_update > 1.0) {
+    std::lock_guard<std::mutex> lck(*(cameras_[index].cloud_msg_mutex_));
```
jkflying: I think we need the mutex at the top, otherwise we could access the timestamp while it is in the middle of being updated and get a corrupted value.

mrivi: true. Fixed

PR #513, opened 2019-10-18
Parameters race condition

mrivi:
Fix race condition between the two thread retrieving the px4 parameters


avoidance/src/avoidance_node.cpp
```diff
@@ -130,13 +132,18 @@ void AvoidanceNode::checkPx4Parameters() {
     }
   };
   while (!should_exit_) {
+    px4_.param_cb_mutex->try_lock();
```
jkflying: Rather make a nested scope here with a lock guard inside.

PR #519, opened 2019-10-25
[WIP] Add Jenkins SITL testing for avoidance

Jaeyoung-Lim:
This PR is adds SITL testing for the avoidance package


TSC21:
@Jaeyoung-Lim can we bring `yaml_cpp_catkin` to the PX4 ROS containers instead? Also, MAVROS is already installed on the Kinetic and Melodic containers.

Jaeyoung-Lim:
Limited build to the avoidance package which doesn't have dependency problems.

@TSC21 yaml_cpp_catkin is not published as a deb package. I would need to see if it is compatible with the `libyaml-cpp-dev`

Jaeyoung-Lim:
@jkflying I would suggest we first merge this as to fix the Jenkins build test failing on master and add the SITL tests in another PR

PR #520, opened 2019-10-25
Fix asynchonous camera transforms

jkflying:
This reworks the condition variables and mutexes around the point cloud transformation. It should now be a) safer and easier to maintain than before b) lower latency and c) lower CPU usage.

It gets rid of the poll for when new transforms arrive in the transform buffer, uses a single mutex for the each camera's point cloud instead of 2, and gets rid of all (I think) non-threadsafe access to the `camera_` vector.

Tested in SITL, needs testing on a real vehicle.

### DIFF
--- a/avoidance/src/transform_buffer.cpp
+++ b/avoidance/src/transform_buffer.cpp
@@ -67,7 +67,7 @@ bool TransformBuffer::getTransform(const std::string& source_frame, const std::s
     return false;
   } else {
     if (iterator->second.back().stamp_ < time) {
-      // "TF Buffer: could not retrieve requested transform from buffer, tf has not yet arrived"
+      print(log_level::warn, "TF Buffer: could not retrieve requested transform from buffer, tf has not yet arrived");
       return false;
     } else if (iterator->second.front().stamp_ > time) {
       print(log_level::warn,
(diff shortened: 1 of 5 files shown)


avoidance/src/transform_buffer.cpp
```diff
@@ -67,7 +67,7 @@ bool TransformBuffer::getTransform(const std::string& source_frame, const std::s
     return false;
   } else {
     if (iterator->second.back().stamp_ < time) {
-      // "TF Buffer: could not retrieve requested transform from buffer, tf has not yet arrived"
+      print(log_level::debug, "TF Buffer: could not retrieve requested transform from buffer, tf has not yet arrived");
```
baumanta: We usually use ROS_DEBUG streams for printing, maybe we should stay consistent.

jkflying: This just links to the ROS_DEBUG below, so I don't think there's an issue.

jkflying:
Tested on a lower-compute platform and found an issue where it would run at 50% update rate if a planning session took slightly over the desired compute time. I've fixed this by decoupling the planner update info and the waypoint generator, it now runs with both waypoint generator at 10Hz and planner at the maximum rate it can, on this platform around 9.2Hz. I've also added some mutexes around other things I noticed might be problematic.

jkflying:
Successfully test flown on iMX8M-mini and Up Core based companion computers.

Up Core flight:
https://logs.px4.io/plot_app?log=8f95bd6b-150c-4d8f-8606-7b4707b5fcf1

Issue #552, opened 2020-01-21
[Global Planner] Unable to cope with RTL & LAND mode


SwiftGust:
**Describe the bug**
Global planner on the master is not able to handle with RTL & Land mode since /mavros/trajectory/desired topic comes in does not contain valid data, no frame_id
while the local planner could work with RTL & Land mode from the firmware.
```
$ rostopic echo mavros/trajectory/desired

---
header: 
  seq: 3574
  stamp: 
    secs: 721
    nsecs: 392112295
  frame_id: "local_origin"
type: 0
point_1: 
  header: 
    seq: 0
    stamp: 
      secs: 0
      nsecs:         0
    frame_id: ''
  coordinate_frame: 0
  type_mask: 0
  position: 
    x: -0.0537663251162
    y: 0.0184620078653
    z: 29.9580383301
  velocity: 
    x: nan
    y: nan
    z: nan
  acceleration_or_force: 
    x: nan
    y: nan
    z: nan
  yaw: nan
  yaw_rate: nan
point_2: 
  header: 
    seq: 0
    stamp: 
      secs: 0
      nsecs:         0
    frame_id: ''
  coordinate_frame: 0
  type_mask: 0
  position: 
    x: -0.0537663251162
    y: 0.0184620078653
    z: 29.9580383301
  velocity: 
    x: nan
    y: nan
    z: nan
  acceleration_or_force: 
    x: nan
    y: nan
    z: nan
  yaw: 2.80641579628
  yaw_rate: nan
point_3: 
  header: 
    seq: 0
    stamp: 
      secs: 0
      nsecs:         0
    frame_id: ''
  coordinate_frame: 0
  type_mask: 0
  position: 
    x: -0.0537663251162
    y: 0.0184620078653
    z: 29.9580383301
  velocity: 
    x: nan
    y: nan
    z: nan
  acceleration_or_force: 
    x: nan
    y: nan
    z: nan
  yaw: nan
  yaw_rate: nan
point_4: 
  header: 
    seq: 0
    stamp: 
      secs: 0
      nsecs:         0
    frame_id: ''
  coordinate_frame: 0
  type_mask: 0
  position: 
    x: nan
    y: nan
    z: nan
  velocity: 
    x: nan
    y: nan
    z: nan
  acceleration_or_force: 
    x: nan
    y: nan
    z: nan
  yaw: nan
  yaw_rate: nan
point_5: 
  header: 
    seq: 0
    stamp: 
      secs: 0
[...]

PR #553, opened 2020-01-21
 [Global Planner] Fix Alt Prior handling

limhyon:
alt_prior_ is pre-defined & limited length std::vector (currently 25 elements are defined)
Accessing out of index of this vector for path finding can cause path finder to fail, thus high altitude RTL is not available in current status.
This would allow high altitude RTL to work in current master, tested in SITL.


SwiftGust:
@Jaeyoung-Lim Fixed code format, but having an issue with coverage.

Jaeyoung-Lim:
Created from https://github.com/PX4/avoidance/issues/552

PR #555, opened 2020-01-23
Improve planning repeatability

jkflying:
Use the longest A* path weighted by cost, since we don't plan all of the way to the goal.

Before, the path chosen could have been any of the green cells, right up to the point of finding the red cell.
image
(play around with it [here](https://qiao.github.io/PathFinding.js/visual/))

Now, it will find roughly the longest path until it actually reaches the end.

Happy for comments on what exactly we can do for a quick win here to improve the choice mechanism for the A* when it only plans partially.

### DIFF
--- a/local_planner/include/local_planner/tree_node.h
+++ b/local_planner/include/local_planner/tree_node.h
@@ -15,6 +15,7 @@ class TreeNode {
   float heuristic_;
   int origin_;
   bool closed_;
+  int depth_;
 
   TreeNode();
   TreeNode(int from, const Eigen::Vector3f& pos, const Eigen::Vector3f& vel);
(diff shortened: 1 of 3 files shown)


jkflying:
SITL test before:
https://logs.px4.io/plot_app?log=323a24ef-8827-4e5e-b6e7-ed24836f3316

After:
https://logs.px4.io/plot_app?log=497302be-9c69-407c-b7e1-dd64dae3eb58

There's definitely an improvement, especially in the occasional odd big 'jump' that doesn't happen anymore.

LorenzMeier:
Is this something that helps short-time?

jkflying:
IMO yes, it needs real-vehicle testing and some more time in SITL though, since it changes the core A* path choice.

baumanta:
I flight tested today, no regression found. Also this PR seems to improve roll and pitch rate
master
master
PR
flight2

PR #557, opened 2020-01-31
Speed Adaption Rework and Cost Function Tuning

baumanta:
This PR adds a lowpass filter to the speed adaption function and restructures it to get rid of discontinuities in the demanded velocity signal. The setpoint coming out of this function undergoes smoothing afterwards, so we thought the very rudimentary velocity calculation would do, but adding this simple filter improved flight behavior significantly.

The plots below show the velocity coming out of the speed adaption function on the right and the signal after smoothing on the left. The derivative shows the commanded acceleration to visualize the difference.

Also I changed the pitching cost slightly, because the drone kept arriving too high at the goal and had to turn and search its way down. Both plots show the same mission with 6 goals in SITL. On the timescale it can be seen that master takes 3:20min to complete whereas the PR takes 2min.

Flight tests pending...

current master:
pr_old

PR:
pr_new

### DIFF
--- a/local_planner/src/nodes/planner_functions.cpp
+++ b/local_planner/src/nodes/planner_functions.cpp
@@ -331,6 +331,7 @@ std::pair<float, float> costFunction(const PolarPoint& candidate_polar, float ob
                                      const Eigen::Vector3f& closest_pt, const bool is_obstacle_facing_goal) {
   // Compute  polar direction to goal and cartesian representation of current direction to evaluate
   const PolarPoint facing_goal = cartesianToPolarHistogram(goal, position);
+  const float goal_distance = (goal - position).norm();
   const Eigen::Vector3f candidate_velocity_cartesian =
       polarHistogramToCartesian(candidate_polar, Eigen::Vector3f(0.0f, 0.0f, 0.0f));
 
@@ -349,8 +350,12 @@ std::pair<float, float> costFunction(const PolarPoint& candidate_polar, float ob
 
   const float yaw_cost = (1.f - weight) * cost_params.yaw_cost_param * angle_diff * angle_diff;
   const float yaw_to_line_cost = weight * cost_params.yaw_cost_param * angle_diff_to_line * angle_diff_to_line;
-  const float pitch_cost =
+  float pitch_cost =
       cost_params.pitch_cost_param * (candidate_polar.e - facing_goal.e) * (candidate_polar.e - facing_goal.e);
+  // increase the pitch cost starting at 5m from the goal (forcing the drone to goal altitude)
+  if (goal_distance < 5.f) {
+    pitch_cost = pitch_cost / ((0.2 * goal_distance) * (0.2 * goal_distance));
+  }
   const float d = cost_params.obstacle_cost_param - obstacle_distance;
   const float distance_cost = obstacle_distance > 0 ? 5000.0f * (1 + d / sqrt(1 + d * d)) : 0.0f;
 
--- a/local_planner/src/nodes/waypoint_generator.cpp
+++ b/local_planner/src/nodes/waypoint_generator.cpp
@@ -338,14 +338,24 @@ void WaypointGenerator::nextSmoothYaw(float dt) {
   setpoint_yaw_rad_ = wrapAngleToPlusMinusPI(setpoint_yaw_rad_);
 }
 
-void WaypointGenerator::adaptSpeed() {
-  speed_ = planner_info_.cruise_velocity;
+void WaypointGenerator::adaptSpeed(float dt) {
+  // lowpass filter the speed to get smoother accelerations
+  const float filter_time_constant = 0.9f;
+  const float alpha = dt / (filter_time_constant + dt);
+  const float last_speed = speed_;
+
+  // at startup parameters are NAN (avoid propagating)
+  if (std::isfinite(planner_info_.cruise_velocity)) {
+    speed_ = planner_info_.cruise_velocity;
+  } else {
+    speed_ = 2.f;
+  }
 
   // If the goal is so close, that the speed-adapted way point would overreach
   float goal_dist = (goal_ - position_).norm();
   if (goal_dist < 1.f) {
-    output_.adapted_goto_position = goal_;
-
+    output_.goto_position = goal_;
+    speed_ = goal_dist;
     // First time we reach this goal, remember the heading
     if (!std::isfinite(heading_at_goal_rad_)) {
       heading_at_goal_rad_ = curr_yaw_rad_;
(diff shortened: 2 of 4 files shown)


LorenzMeier:
Nice! As we are concerned with pitch angular rate, could you also plot that going forward? Because that is the real measure of how smooth a flight looks. Do you have a plot of that?

baumanta:
here the topics regarding pitch from the ulog (same mission in SITL):
master:
pitch_old
PR:
pitch_new

local_planner/src/nodes/waypoint_generator.cpp
```diff
@@ -339,13 +339,23 @@ void WaypointGenerator::nextSmoothYaw(float dt) {
 }
 
 void WaypointGenerator::adaptSpeed() {
-  speed_ = planner_info_.cruise_velocity;
+
+  //lowpass filter the speed to get smoother accelerations
+  float lp_filter_constant = 0.1;
```
Jaeyoung-Lim: Is this a value that is general enough that we can hard code in the code?

local_planner/src/nodes/waypoint_generator.cpp
```diff
@@ -339,13 +339,23 @@ void WaypointGenerator::nextSmoothYaw(float dt) {
...
+    speed_ = 2.f;
+  }
 
   // If the goal is so close, that the speed-adapted way point would overreach
   float goal_dist = (goal_ - position_).norm();
   if (goal_dist < 1.f) {
-    output_.adapted_goto_position = goal_;
-
+    output_.goto_position = goal_;
+    speed_ = goal_dist;
```
jkflying: This could give you a step in the `speed_` when you hit this radius.

local_planner/src/nodes/planner_functions.cpp
```diff
@@ -350,7 +351,7 @@ std::pair<float, float> costFunction(const PolarPoint& candidate_polar, float ob
   const float yaw_cost = (1.f - weight) * cost_params.yaw_cost_param * angle_diff * angle_diff;
   const float yaw_to_line_cost = weight * cost_params.yaw_cost_param * angle_diff_to_line * angle_diff_to_line;
   const float pitch_cost =
-      cost_params.pitch_cost_param * (candidate_polar.e - facing_goal.e) * (candidate_polar.e - facing_goal.e);
+      cost_params.pitch_cost_param * (candidate_polar.e - facing_goal.e) * (candidate_polar.e - facing_goal.e) / (0.2 * goal_distance);
```
Jaeyoung-Lim: Where does the 0.2 come from?

local_planner/src/nodes/waypoint_generator.cpp
Jaeyoung-Lim: Could you explain the rational behind defining the goal distance to speed directly?

local_planner/src/nodes/waypoint_generator.cpp
baumanta: I agree that having hard coded number is not ideal, but I don't think that we will ever need to touch that one again, so I don't think it makes sense to make it a parameter

local_planner/src/nodes/planner_functions.cpp
baumanta: so before the pith cost was independent from the distance to the goal and this resulted in the drone always arriving too high, because the pitch cost was way too high in comparison to avoiding the ground (like 6m below). With this addition the pitch terms starts being lower than before at 5m from the goal, going to zero the closer we get. This allows the drone to actually reach the point. Same here, I don't think we need to tough this number again

local_planner/src/nodes/waypoint_generator.cpp
baumanta: The speed_  variable defines the distance of the setpoint from the current position of the drone. It is not speed in the general sense though, not in m/s at least. As we use the setpoint as a carrot-stick thing in the firmware, father means faster, closer means slower. So setting the setpoint in goal direction at the distance goal_dist basically puts the setpoint onto the goal.

local_planner/src/nodes/planner_functions.cpp
Jaeyoung-Lim: Thanks for the explanation. I am just a bit uncomfortable of multiplying some number without any explanation and without it being exposed. Could you maybe add a comment on what 0.2 means and why? I think that would be already helpful

local_planner/src/nodes/waypoint_generator.cpp
```diff
@@ -360,16 +370,19 @@ void WaypointGenerator::adaptSpeed() {
...
-    // Scale the pose_to_wp by the speed
-    Eigen::Vector3f pose_to_wp = output_.goto_position - position_;
-    if (pose_to_wp.norm() > 0.1f) pose_to_wp.normalize();
-    pose_to_wp *= std::min(speed_, goal_dist);
-
     heading_at_goal_rad_ = NAN;
-    output_.adapted_goto_position = position_ + pose_to_wp;
   }
 
+  speed_ = lp_filter_constant * speed_ + (1.f - lp_filter_constant) * last_speed;
```
bresch: Be aware that the real time constant of your filter will change if dt changes. 
A safer implementation is:
```
alpha = dt / (time_constant + dt)
out = alpha * in + (1 - alpha) * out_prev
```
if the `time constant >> dt`, then you can simplify the formula by  `alpha = dt * time_constant_inverse` and give `time_constant_inverse` as a `static constexpr`

local_planner/src/nodes/waypoint_generator.cpp
baumanta: the lowpass filter comes after that, so I think it should be fine, no?

jkflying: Ah, I misread the way your filter works. Typically you have one variable as the state, which gets updated gradually. Your filter instead 'clamps' your speed back to 90% of the previous speed. It works, it just isn't idiomatic.

It think it would be better to re-implement it clearly in terms of inputs and outputs, using the looprate-invariant formula from @bresch below.

baumanta:
Flight on the mule: https://suite.auterion.com/flights/3bc06e83-3c9f-4258-9f73-f671d62e46cb
Pitch rate setpoint signal seems to have slightly smaller amplitude. But the mule always flew nice enough. As we had the main problem on the alta I would like to see it fly there as well

this PR
mule_new

older log as comparison:
mule_old

LorenzMeier:
@baumanta This is nitpicking, but an important one: Given the data the mule actually didn't "fly nice", but the issue wasn't as visible. It is a big difference if you can just not see it on a platform or if it is indeed not there.

What this also says is that the current mule configuration is not suitable for testing, because it is hiding the issue.

baumanta:
Logs from the other drone confirm that pitch rate goes down a lot:
this PR:
alta_pitch_new
master:
alta_pitch_old

PR #560, opened 2020-02-21
add avoidance in loiter mode

mrivi:


baumanta:
Do I get this right, this will make the planner do the calculations also in loiter mode. But on the firmware side it will still not be used, right?

jkflying:
It needs a change in the firmware, this will be coming soon.

PR #579, opened 2020-05-17
Clarify Ubuntu 18.04 is preferred.

hamishwillee:


hamishwillee:
PS @jkflying With PX4 moved to 18.04 should we be even stronger and remove the 16.04 version?

jkflying:
Removing mention of 16.04 and replacing all mentions of "kinetic" with "melodic" would be great, yes.

hamishwillee:
@jkflying There you go then.

PR #590, opened 2020-07-01
Use C++ 14 instead of C++ 11 (PCL 1.10 requires C++14 as the minimum)

nikhilprakash99:
catkin build error in Focal ( Ubuntu 20.04 ) due to PCL dependancy
Latest PCL (1.10) requires C++14 as the minimum


TSC21:
@nikhilprakash99 did you already validated this? Is it working in Ubuntu Focal with ROS Noetic?

PR #599, opened 2020-08-07
Fix build warnings

mortenfyhn:
I tried to build the avoidance repo just now, and the build failed
with clang. The error in clang is just a warning in gcc, so gcc builds,
but with warnings. I've taken the warnings reported by both clang and
gcc, and tried to fix them.

The issues were

* assigning from an uninitialized variable,
* reaching the end of a non-void function, and
* passing a std::string to a "%s" specifier (expects char*).


jkflying:
Thanks! The global planner hasn't had much attention so I suspect there's a lot more of these kind of things...

PR #601, opened 2020-08-13
Risk based speedup algorithm added

ldg810:
In global planner, the setpoint to follow global path is updated periodically at "GlobalPlannerNode::publishSetpoint()"
The speed of drone is keep constantly (2.0 m/s at current situation).
However, it would be efficient to speed up in low-risk environment.

I added following two parameters to apply risk-based speedup algorithm.
- max_speed_(double) : maximum speed of drone in safe environment.
- use_risk_based_speedup_(bool) : boolean to use risk based speedup algorithm

The algorithm is simple
1. get current cell from global_planner
2. get risk of current cell (=cur_risk)
3. if risk is high(>=0.5), set speed to slow (1.0m/s)
    else, set speed to fast (1.0 + (global_planner_.max_speed_ - 1.0) * (1 - cur_risk))
4. isCloseToGoal is also changed to determine the drone is close to goal if distance from destination and drone is within current speed setpoint. (If not, the drone would oscillate in fast flight)

Following graph is comparison between default algorithm(red line) and risk based speedup algorithm(green line) from a point to another point
abcd

PR #603, opened 2020-08-26
Parameterize camera frame

ldg810:
To use global_planner with real drone system, frame transformation between fcu and camera should work well.

In current code, camera_link is set as default camera frame id but [PX4 Vision Kit](https://docs.px4.io/master/en/complete_vehicles/px4_vision_kit.html) using camera_main_map as figure below.

PX4VisionKit-tf map

I think the camera frame should be parameterized in launch file for general use of global_planner.

### DIFF
--- a/global_planner/include/global_planner/global_planner_node.h
+++ b/global_planner/include/global_planner/global_planner_node.h
@@ -110,6 +110,7 @@ class GlobalPlannerNode {
   double start_yaw_;
   bool position_received_;
   std::string frame_id_;
+  std::string camera_frame_id_;
 
   // Dynamic Reconfiguration
   double clicked_goal_alt_;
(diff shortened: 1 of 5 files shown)


jkflying (approved):
Looks good, thanks for the contribution!

I guess the speed change is to make it safe for real vehicles, if you want maybe follow up with a PR to parameterize it?

ldg810:
> Looks good, thanks for the contribution!
> 
> I guess the speed change is to make it safe for real vehicles, if you want maybe follow up with a PR to parameterize it?

Yes, speed change is to make it safe for real vehicles.

Parameterizing default speed is included in another [PR #601](https://github.com/PX4/avoidance/pull/601).

I will test it with a real vehicle and share results on there.

PR #611, opened 2020-11-27
improve avoidance node to use ROS2

ldg810:
I am working on to use PX4-Avoidance global_planner in ROS2 environment.

Until now, I succeed with subscribing a topic(VehicleLocalPosition) in ROS2 environment.

The avoidance node should be changed to the following reasons.
- In current form, the two avoidance nodes(avoidance_node_cmd, avoidance_node_status) spinning and it stuck with executor.join() function. Thus, I made thread/related variables as global variable to make it spin in the background.
- tf2::fromMsg usage was fault. I change the order of input parameters but I failed to use "tf2::fromMsg(tf_earlier.transform.translation, tf_earlier_translation);". In tf2 documentation, there is a declaration about geometry_msgs::msg::Vector3 and tf2::Vector3 but it is not working.. So I changed it to "tf_earlier_translation.setValue(tf_earlier.transform.translation.x, tf_earlier.transform.translation.y, tf_earlier.transform.translation.z);" as directly assign Vector3 values to tf_earlier_translation. (Please give a comment about this)

The message type of mavros and px4_msgs in ROS2 is quite different so I thinks its little bit hard work to porting global_planner to ROS2 environment but I will try to solve it.

PR #617, opened 2020-12-22
global_planner for ROS2 initial version

ldg810:
This PR includes minimal re-making existing global_planner for the purpose of applying ROS2.

Now, we can build (using "colcon build") and start global_planner using binary directly.

Many things are remained to do including..
- Make config file for parameters.
- Include PathWithRiskMsg.msg (There are difficulty to include .msg in same package such as [find_package(PCL) interferes with message generation issue](https://github.com/ros2/rosidl/issues/402).
- Validate tf2 transformation working correctly
- Enable information log for library cpp files.
- Make launch file for launching px4 and avoidance module at the same time.

**- This update should be applied after PR #611.**


ldg810:
The ros2 version of global planner was improved a lot in my private repository.

I will refactor code and make a new PR.

PR #631, opened 2021-03-08
Fix local_planner config path after #483

bys1123:
`local_planner/resource/px4_config.yaml` already moved to `avoidance/resource/px4_config.yaml` in #483 but there is there config files forgot to change the path. Here is the fix.


bys1123:
@Jaeyoung-Lim OK, added. thanks for review.

Jaeyoung-Lim (approved):
LGTM!

PR #692, opened 2023-01-04
Update Readme for Ubuntu 20.04 and fix PX4 changes

jhiggason:
While trying to get a working simulator going- I found that the only way to do so with the current packages available was to switch to Ubuntu 20.04 and ROS Noetic.  I also updated the new path for setup_gazebo.bash and changed some of the instructions to more match what is on ROS Noetic's install page.  

I've only ever been able to get this working on a desktop version of 20.04- For whatever reason- virtual machines are not functioning the same.  Once i figure that portion out I will make notes in the readme on how to make a virtual machine work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,6 +193,7 @@ In the following section we guide you through installing and running a Gazebo si
    ```
 
 The last three steps, together with sourcing your catkin **setup.bash** (`source ~/catkin_ws/devel/setup.bash`) should be repeated each time a new terminal window is open.
+
 OR you can ensure your ~/.bashrc is setup correctly so that each new terminal is sourced correctly.  
 
 ```bash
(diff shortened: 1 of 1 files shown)


julianoes:
FYI @baumanta

nicovanduijn (approved):
Thank you for your contribution. I suppose we should also update the CI at some point, but that's a story for another day...

Jaeyoung-Lim:
@nicovanduijn Actually we already are testing against ROS Noetic: https://github.com/PX4/PX4-Avoidance/blob/79c2f3b28b6ae48fdbedc46eebaeea043b12e556/.github/workflows/build_test.yml#L18

nicovanduijn:
@Jaeyoung-Lim not everywhere, right? I guess what's broken is this:
https://github.com/PX4/PX4-Avoidance/blob/79c2f3b28b6ae48fdbedc46eebaeea043b12e556/.github/workflows/valgrind_test.yml#L17

Jaeyoung-Lim:
@nicovanduijn I think the Valgrind tests were supposed to be ros distribution agnostic - As far as I understand, this is why not even melodic was being tested there.

